PR #23, opened 2026-04-15
fix: sanitize host-based emitter routing keys to block path traversal

DavidJBianco:
### Motivation

- Close a path traversal/arbitrary file-write vector where scenario-supplied domain/FQDN strings (e.g., `proxy_fqdn`) were used directly as filesystem path components in host-multiplexed emitters.

### Description

- Sanitize routing keys in `HostMultiplexEmitter._get_writer()` by calling a new `HostMultiplexEmitter._sanitize_host_fqdn()` helper and key the writer cache by the sanitized value instead of the raw input.
- Implement `_sanitize_host_fqdn()` to replace unsafe characters, collapse `..` sequences, trim leading/trailing dots/spaces, and fall back to `unknown-host` for empty/invalid results.
- Update path construction to use the sanitized host directory when creating per-host writer paths.
- Add a unit test `tests/unit/test_host_base_sanitization.py` that attempts to route to `"../../.ssh"` and asserts no files are written outside the configured output directory, and update `TODO.md` to reflect the security fix.

### Testing

- Ran linter/format checks with `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.
- Added a focused unit test for path sanitization; attempted to run `PYTHONPATH=src uv run pytest tests/unit/test_host_base_sanitization.py -q` but execution in this environment failed due to a missing test dependency (`yaml`), so the test could not be executed here.
- Changes were validated locally via code inspection and an automated test was added to enforce the behavior in CI where full dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e017a5004483258eef31a282d37a1b)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -82,6 +82,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 - [x] **CI runs tests 3 times** — 3 separate pytest invocations (unit, integration, both again for coverage). Consolidate to single run.
 - [x] **No pre-commit hooks** — ruff issues only caught in CI. Add pre-commit framework with ruff check + format hooks.
 - [ ] **Re-generation appends to existing output** — `GenerationEngine` creates output directories with `exist_ok=True` and emitters append to existing files. Re-running a scenario without manually clearing the output directory produces mixed old+new data. Should clean the output directory (or at least its per-sensor subdirectories) before writing.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing — this vulnerability is already fixed on `dev` by commit 7ef6270 ("security: shared path safety utility + fix 3 unprotected emitters"). That commit creates a shared `sanitize_path_component()` utility in `src/evidenceforge/utils/paths.py` and applies it to all 5 vulnerable emitters (host_base, windows, sysmon, zeek_base, bash_history), with 20 unit tests. Will be merged to main via PR shortly.

PR #34, opened 2026-04-15
fix(ecar): harden EcarEmitter._render_event against malformed raw payloads

DavidJBianco:
### Motivation

- Prevent untrusted/malformed `raw` eCAR payloads from crashing generation when required keys are missing, which previously caused `KeyError` in `EcarEmitter._render_event`.

### Description

- Use safe accessors in `EcarEmitter._render_event()` so `timestamp`, `object`, and `action` default to benign values and numeric timestamps are supported; `timestamp_ms` now falls back to `0`, and `object`/`action` default to `"UNKNOWN"` when absent.
- Preserve existing eCAR spec behavior for valid inputs (datetime and numeric timestamps, pid/tid/ppid handling, property stringification) and keep `actorID`/`objectID` logic unchanged.
- Add unit tests exercising malformed raw payloads and numeric timestamp handling, and update `TODO.md` to mark the remediation complete.

### Testing

- Ran linting and formatting checks: `uv run ruff check .` passed and `uv run ruff format --check .` passed after formatting.
- Added targeted tests in `tests/unit/test_ecar_spec_compliance.py`; executing the test file in this environment failed due to a missing runtime dependency (`yaml`) in the container, but a direct smoke test using `PYTHONPATH=src` invoked `_render_event()` and verified the fallback behavior (passed).
- Verified repository state and updated `TODO.md` to record the fix as completed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e017a4b6ec8325afb5bd3028179cd8)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -85,6 +85,8 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Tier 1: Foundational Correctness
 
+- [x] eCAR raw events missing required keys no longer crash generation — eCAR renderer now defaults missing timestamp/object/action to safe values for malformed raw payloads.
+
 Data is *wrong* — a hunter hits dead ends. Fix these first; several unblock Tier 2 work.
 
 - [x] **LogonIDs leak across hosts** — remote processes on DC/file server use the originating-host LogonID instead of the destination host's 4624 TargetLogonId. Breaks every pivot-based hunting workflow.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing — the `_render_event` KeyError risk is real but low-severity: all internal callers (logon, process, connection, etc.) set `timestamp`, `object`, and `action` fields. The `emit_raw()` path is an escape hatch that intentionally assumes complete data. Adding silent fallbacks like `action="UNKNOWN"` would produce silently wrong eCAR records that pass validation but contain meaningless data — it's better to crash loudly on truly malformed input so the caller knows to fix their payload.

PR #36, opened 2026-04-15
fix: harden evaluation parser discovery against symlink traversal

DavidJBianco:
### Motivation
- The evaluator's `discover_log_files()` previously descended into per-sensor subdirectories and accepted files via `Path.is_file()`/`is_dir()` without symlink or resolved-path containment checks, allowing symlinked sensor dirs/files to reference arbitrary files outside the evaluation `output_dir`.
- The intent is to prevent accidental or malicious arbitrary file reads during evaluation while preserving the existing file-discovery semantics for legitimate logs under the output directory.

### Description
- Add `output_root = output_dir.resolve()` and helper guards `def _is_safe_directory(path: Path) -> bool` and `def _is_safe_file(path: Path) -> bool` that reject symlinks and require `path.resolve().relative_to(output_root)` containment before accepting entries.
- Replace raw `is_file()`/`is_dir()` checks in `discover_log_files()` with the safe helpers and apply the safety checks to top-level files, per-host/per-sensor directories, nested bash_history discovery, and deeper per-sensor traversal.
- Add unit tests `test_skips_symlinked_sensor_directories` and `test_skips_symlinked_top_level_files` to `tests/unit/test_eval_parsers.py` to validate symlinked dirs and symlinked top-level files are ignored, and update `TODO.md` to mark the security hardening task completed.

### Testing
- Ran linter and formatter checks with `uv run ruff check .` and `uv run ruff format --check .`, which passed after formatting adjustments.
- Reformatted affected files with `uv run ruff format` and committed the changes, and added the new tests under `tests/unit/test_eval_parsers.py`.
- Attempted to run the unit tests with `PYTHONPATH=src uv run pytest tests/unit/test_eval_parsers.py`, but test execution could not complete in this environment due to a missing runtime dependency (`PyYAML` / `yaml`); the tests exercise the new symlink-safety behavior and should run successfully in CI or a developer environment with project dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e017a38a308325aa739049b3c0e234)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -75,6 +75,9 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Tier 0: Infrastructure
 
+- [x] Security: hardened evaluation log discovery against symlink/path traversal in per-sensor directories (reject symlinked dirs/files and enforce resolved-path containment under output_dir).
+
+
 - [x] Security: blocked symlinked `eforge` install target in `install-skills` to prevent arbitrary overwrite/deletion and stale-file cleanup outside target.
 
 - [x] **`uv.lock` not committed** — gitignored, so CI `setup-uv@v4` cache fails. Remove from `.gitignore` and commit.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing — the symlink traversal bug in `discover_log_files()` is real and we'll fix it, but using the shared `reject_symlink()` utility from `src/evidenceforge/utils/paths.py` instead of reinventing inline helpers. This keeps symlink rejection consistent across the codebase (emitters, install-skills, overlay, and now eval parser discovery all use the same utility). Will apply the fix on `dev` shortly.

PR #39, opened 2026-04-16
fix: prevent dict-valued bash command pools from persona selection

DavidJBianco:
### Motivation
- Prevent a denial-of-service crash where `_get_role_pool()` could return non-command YAML sections (e.g., `params` or `keyboard_adjacency`) which then flow to `random.choice()` and raise `KeyError` when generation is run with a crafted persona name.

### Description
- Constrain exact-match persona lookup in `_get_role_pool()` to only return keys whose YAML value is a list (`isinstance(data[persona_lower], list)`), add a unit test `TestBashRolePoolSelection.test_non_list_yaml_sections_are_not_used_as_role_pool` asserting persona `params` falls back to `sysadmin`, and update `TODO.md` to mark the security fix complete.

### Testing
- Ran lint/format checks: `uv run ruff check .` passed and `uv run ruff format --check .` passed; added unit test is present under `tests/unit/test_expert_round4.py`; running `PYTHONPATH=src uv run pytest tests/unit/test_expert_round4.py -q` in this environment failed to start due to a missing optional dependency (`yaml`) causing an import error, not due to the change itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e022aa008c8325bffa1725862c199e)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -75,6 +75,8 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Tier 0: Infrastructure
 
+- [x] Security: restricted custom persona bash command pools to list-valued entries to avoid dict-key DoS crash in `pick_bash_command()`; added regression test for persona name `params`.
+
 - [x] Security: blocked symlinked `eforge` install target in `install-skills` to prevent arbitrary overwrite/deletion and stale-file cleanup outside target.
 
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing — the bug is not exploitable in practice. The `_get_role_pool()` return value flows through `commands.get(pool_key, commands.get("common", ["ls"]))` in `pick_bash_command()`, which provides a safe fallback. The YAML data structure doesn't contain persona-named dict keys that would match the exact-match path. The `isinstance(data[key], list)` check is technically correct but unnecessary given the defensive call site.

PR #53, opened 2026-04-17
fix: add ProcessAccessEventSpec to scenario EventSpec union

DavidJBianco:
### Motivation
- Validation treats `process_access` as a Windows-only event but the scenario model lacked a corresponding typed spec, causing Pydantic to reject storyline YAML that uses `process_access`.

### Description
- Add a `ProcessAccessEventSpec` model (`type: "process_access"`) with `target_process` and `access_mask` fields and include it in the `EventSpec` discriminated union so `process_access` tags can be parsed.
- Add a unit test `test_storyline_event_with_process_access_event` that constructs a `StorylineEvent` containing a `process_access` event and update `TODO.md` to mark the issue resolved.

### Testing
- Ran `uv run ruff check .` and formatted changed files with `uv run ruff format`, and linter/format checks passed after formatting.
- Attempted `PYTHONPATH=src uv run pytest tests/unit/test_models.py -k "process_access or storyline_event_with_events_list"` but pytest collection failed in this environment due to a missing `yaml` dependency, so the new test was added but not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e0c61922308325837e184e674cb293)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -2,7 +2,7 @@
 
 **Status:** Phase 8.5 (Dual src/dst HostContext) COMPLETE; Pre-MVP quality fixes ongoing
 **Started:** 2026-03-11
-**Last Updated:** 2026-04-08
+**Last Updated:** 2026-04-16
 
 See [CHANGELOG.md](CHANGELOG.md) for detailed development history of completed phases.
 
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing as duplicate of #52. Both address the same issue (ProcessAccessEventSpec missing from EventSpec union). #52 is the canonical PR.

PR #43, opened 2026-04-16
fix: avoid IPv6 crash in DNS AAAA fallback (preserve IPv6 literals)

DavidJBianco:
### Motivation
- The AAAA DNS answer path previously used `_IPV6_MAP.get(dst_ip, _ipv4_to_fake_ipv6(dst_ip))`, which eagerly calls `_ipv4_to_fake_ipv6()` and raises `ValueError` when `dst_ip` is a valid IPv6 literal, causing a denial-of-service during generation.
- The intent is to ensure DNS AAAA answers are produced deterministically without crashing when scenarios include IPv6 addresses.

### Description
- Changed AAAA answer selection in `ActivityGenerator._emit_dns_lookup()` so it: uses `_IPV6_MAP` when present, preserves an unmapped IPv6 literal (`':' in dst_ip`), and only calls `_ipv4_to_fake_ipv6(dst_ip)` for IPv4 destinations (`src/evidenceforge/generation/activity/generator.py`).
- Added a regression unit test `test_aaaa_query_handles_ipv6_destination` to `tests/unit/test_phase5_network_diversity.py` that forces the AAAA branch for an IPv6 `dst_ip` and asserts the DNS event emits successfully with the IPv6 answer.
- Updated `TODO.md` to record this security fix per project workflow.

### Testing
- Ran lint checks: `uv run ruff check .` succeeded.
- Ran format checks: `uv run ruff format --check .` succeeded.
- Attempted to run the focused unit test suite with `PYTHONPATH=src uv run pytest tests/unit/test_phase5_network_diversity.py -k "aaaa"`, but the environment is missing runtime dependencies (`yaml`), causing an import-time failure before the tests could execute; the regression test was added and is ready to run in CI or a fully provisioned dev environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e022a9028c8325bd624ea607a27312)

PR #54, opened 2026-04-17
fix: route Linux logon syslog events to destination host

DavidJBianco:
### Motivation
- Correct a regression in the dual `src_host`/`dst_host` model where Linux `logon` SyslogContext (attached to the destination host) could be routed to the source host because the emitter previously preferred `src_host` first.

### Description
- Update `SyslogEmitter._linux_host()` to prefer `dst_host` for `logon` events, add a focused unit test verifying Linux→Linux `logon` routing to the destination host, and mark the fix as completed in `TODO.md`.

### Testing
- Ran static checks: `uv run ruff check .` and `uv run ruff format --check .` (both passed); a new unit test was added but running `PYTHONPATH=src uv run pytest tests/unit/test_dispatcher.py -k syslog_linux_host_prefers_destination_for_logon` was blocked in this environment due to a missing dependency (`yaml` / PyYAML) so the test could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e0c619577c8325aaad16793e1ba695)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -47,6 +47,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Recently Resolved
 
+- [x] Fix syslog host routing for Linux↔Linux network logons (ensure destination host receives sshd syslog entry)
 - [x] Fix `_find_user_session` mixed tz-aware/naive `start_time` comparison crash (Aardvark finding)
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Already fixed on dev. The dst_host routing for Linux logon syslog events was implemented in commit 90fdd2b (Dual src/dst HostContext phase, PR #43).

PR #57, opened 2026-04-17
fix: harden Windows eval XML parser against entity expansion DoS

DavidJBianco:
### Motivation
- The Windows Event XML parser used `ElementTree.fromstring()` on untrusted event fragments and previously read entire files into memory, enabling entity/DTD-based XML DoS and large-file memory use.
- The change prevents attacker-controlled XML (e.g., nested entity expansions) from being expanded or from causing high memory usage during evaluation.

### Description
- Stream `windows_event_security.xml` line-by-line and extract individual `<Event>` elements without reading the whole file by replacing the global regex/full-file read with `parse_file()` that accumulates lines between `EVENT_START_PATTERN` and `EVENT_END_PATTERN`.
- Detect and explicitly reject unsafe XML declarations by checking for `<!DOCTYPE` or `<!ENTITY` in the extracted fragment before parsing and returning a parse error instead of calling `ElementTree.fromstring()` on dangerous content.
- Add two focused unit tests in `tests/unit/test_eval_parsers.py` to assert rejection of DOCTYPE/entity declarations and to ensure `parse_file()` streams input (does not call `Path.read_text()`).
- Update `TODO.md` to mark the security hardening item completed.

### Testing
- Ran linting and formatting checks: `uv run ruff check src/evidenceforge/evaluation/parsers/windows.py tests/unit/test_eval_parsers.py` passed and `uv run ruff format --check ...` passed after formatting changes.
- Added unit tests `test_rejects_doctype_and_entity_declarations` and `test_parse_file_streams_without_reading_full_content` under `tests/unit/test_eval_parsers.py` (tests are present but full `pytest` runs were blocked in this environment due to missing runtime dependencies and network-restricted package sync).
- Attempted to run `uv run pytest tests/unit/test_eval_parsers.py` but execution was prevented by missing runtime packages in the environment (e.g., `pydantic`, `pyyaml`), so test execution is pending in CI/normal dev environments where dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e0c61ac0c48325bf415f42ebbd79e2)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -47,6 +47,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Recently Resolved
 
+- [x] Security: harden eval Windows XML parser against untrusted entity expansion and large-file memory usage
 - [x] Fix `_find_user_session` mixed tz-aware/naive `start_time` comparison crash (Aardvark finding)
 - [x] Baseline inbound profile traffic no longer depends on outbound role traffic for business-hour gating (fixed UnboundLocalError when outbound profile is empty).
 - [x] Security: validate blocked_c2 interval/duration are > 0 to prevent zero-interval infinite loop DoS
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Consolidated into PR #63 (merged to dev). Cherry-picked with tests.

PR #58, opened 2026-04-17
fix: enforce JSON Logic truthiness for field constraints

DavidJBianco:
### Motivation

- JSON Logic field constraints were only treated as failures when the evaluator returned literal `False` or `None`, which allowed falsey non-boolean results (`0`, `""`, `[]`) to bypass validation.

### Description

- Change `validate_field_constraints` in `src/evidenceforge/formats/validator.py` to treat any falsey JSON Logic result as a validation failure by using `if not logic_result`.
- Add a regression unit test in `tests/unit/test_format_validator.py` that parametrizes falsey non-boolean values (`0`, `""`, `[]`) and asserts they are rejected by `validate_field_constraints`.
- Update `TODO.md` to mark the JSON Logic truthiness validation fix as completed.

### Testing

- Lint/format checks were run with `uv run ruff check .` and `uv run ruff format --check .` and passed.
- A quick functional check using `PYTHONPATH=src uv run python -c "from evidenceforge.formats.validator import validate_field_constraints; ..."` confirmed the validator now rejects a falsey non-boolean JSON Logic result (`False` printed for the test invocation).
- Attempting to run the full unit test invocation in this environment (`PYTHONPATH=src uv run pytest tests/unit/test_format_validator.py`) was blocked by a missing runtime dependency (`ModuleNotFoundError: No module named 'yaml'`), so the broader test suite could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e0c619121883259a0f970b1bcea250)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -47,6 +47,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Recently Resolved
 
+- [x] Security: restored JSON Logic field-constraint truthiness validation so falsey non-boolean outputs (0, "", []) fail validation
 - [x] Fix `_find_user_session` mixed tz-aware/naive `start_time` comparison crash (Aardvark finding)
 - [x] Baseline inbound profile traffic no longer depends on outbound role traffic for business-hour gating (fixed UnboundLocalError when outbound profile is empty).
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Consolidated into PR #63 (merged to dev). Cherry-picked with tests.

PR #62, opened 2026-04-17
fix: prevent DLL profile loader crash on malformed loaded_modules entries

DavidJBianco:
### Motivation

- The DLL profile loader assumed every `loaded_modules` list item was a mapping and accessed fields with `entry.get(...)` / `entry["path"]`, which can raise `AttributeError`/`TypeError` when overlays contain non-dict items and crash generation.

### Description

- Harden `src/evidenceforge/generation/activity/dll_load_profiles.py` by making `_validate_entry` accept `entry: Any` and early-return `False` with a logged warning when `entry` is not a mapping or when `path` is not a string, preventing attribute access on non-dict values.
- Kept default-application behavior in `_apply_defaults` unchanged and preserved merging/duplicate-detection logic in `load_dll_profiles()` so runtime behavior is unaffected for valid inputs.
- Added unit tests in `tests/unit/test_dll_load_profiles.py` covering non-mapping entries and non-string `path` values to verify the loader now skips malformed entries safely.
- Updated `TODO.md` per project workflow to record the security hardening task as completed.

### Testing

- Ran `uv run ruff check .` which passed successfully.
- Ran `uv run ruff format --check .` which reported files are formatted (no failures).
- Attempted `uv run pytest tests/unit/test_dll_load_profiles.py` but the container lacked runtime dependencies (`ModuleNotFoundError: yaml`), causing the local pytest run to fail; the new unit tests were added and are expected to pass in a correctly provisioned CI environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e2549add588325bd8f19ced9693083)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -47,6 +47,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Recently Resolved
 
+- [x] Security: harden DLL profile loader against non-dict loaded_modules overlay entries (prevent AttributeError/TypeError DoS)
 - [x] Fix `_find_user_session` mixed tz-aware/naive `start_time` comparison crash (Aardvark finding)
 - [x] Baseline inbound profile traffic no longer depends on outbound role traffic for business-hour gating (fixed UnboundLocalError when outbound profile is empty).
 - [x] Security: validate blocked_c2 interval/duration are > 0 to prevent zero-interval infinite loop DoS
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Already fixed on dev. The _validate_entry() function with graceful error handling for malformed loaded_modules entries was included in commit 4e2edc3 (data-driven DLL load profiles for Sysmon Event 7).

PR #81, opened 2026-04-26
fix: preserve storyline monotonic ordering with typing cadence offsets

DavidJBianco:
### Motivation
- Storyline steps that contain multiple intra-step events were using a per-step timestamp guard that ignored per-event typing cadence offsets, allowing subsequent steps to be scheduled before the last intra-step event and producing non-monotonic timeline/state regressions.
- The change ensures timeline monotonicity across storyline steps when `typing_cadence()` adds cumulative intra-step delays.

### Description
- Advance the step-level `_prev_event_time` to the final intra-step event timestamp (step base time + `cadence_offsets[-1]`) after executing a compound step so later steps cannot be scheduled earlier; change applied in `src/evidenceforge/generation/engine/storyline.py`.
- Add a deterministic regression test `test_execute_storyline_uses_last_intra_step_timestamp_for_monotonic_ordering` in `tests/unit/test_logonid_scoping.py` that simulates two adjacent steps and verifies monotonic ordering when cadence introduces multi-second offsets.
- Update `TODO.md` to mark the storyline typing-cadence monotonic time guard fix as resolved.

### Testing
- Ran linter/format checks: `uv run ruff check .` succeeded and `uv run ruff format` reformatted the new test file, after which `uv run ruff format --check .` passed.
- Attempted unit test run with `PYTHONPATH=src uv run pytest tests/unit/test_logonid_scoping.py -q`, but the test run could not complete in this environment due to missing runtime dependency `PyYAML` (`ModuleNotFoundError: No module named 'yaml'`), so full pytest was not executed here; the added test is deterministic and should pass in CI where dependencies are available.
- The change is low-risk and narrowly scoped to storyline scheduling bookkeeping and test coverage for the regression.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e0c61aa4ec832596a668c086b05d95)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -47,6 +47,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 
 ### Recently Resolved
 
+- [x] Fix storyline typing cadence monotonic time guard (`_prev_event_time`) so subsequent steps cannot move backward relative to intra-step events
 - [x] Fix `_find_user_session` mixed tz-aware/naive `start_time` comparison crash (Aardvark finding)
 - [x] Baseline inbound profile traffic no longer depends on outbound role traffic for business-hour gating (fixed UnboundLocalError when outbound profile is empty).
 - [x] Security: validate blocked_c2 interval/duration are > 0 to prevent zero-interval infinite loop DoS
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Superseded by dev commit 1b180a3, which reapplies the storyline typing-cadence monotonicity fix and regression test on the current dev line.

PR #120, opened 2026-05-11
fix: bound Windows emitter event buffering

DavidJBianco:
### Motivation
- The Windows event emitter deferred rendering until close and no longer flushed on `buffer_size` or barrier events, allowing untrusted scenarios to accumulate O(total events) `_event_dicts` in memory and enabling local memory DoS.\

### Description
- Implement an on-disk chronological spool (SQLite) for deferred Windows event dictionaries and track spool state alongside the in-memory buffer so in-memory retention is bounded at `buffer_size`/barrier time.\
- Spill buffered event dicts to the spool at non-threaded `buffer_size` thresholds and during threaded queue/barrier flush boundaries instead of retaining them in `_event_dicts`.\
- Preserve final chronological rendering and per-host `EventRecordID` assignment by streaming spooled events in timestamp+sequence order during the final `_flush_unlocked()`/close, then clean up the temporary spool.\
- Add regression tests covering non-threaded buffer spooling and threaded barrier spooling, and update `TODO.md` to record the security fix.\

### Testing
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.\
- Ran focused unit tests for the Windows emitter and threading (`pytest` targets) which passed locally.\
- Ran the full unit/integration suite with `UV_PROJECT_ENVIRONMENT=.venv uv run pytest --no-cov -q`, which completed successfully (`2658 passed, 37 skipped`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a01ebb64fbc8325ba55524cc68a8978)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -48,6 +48,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 - [x] Fix causal-ordering scoring so diagnostic sample caps do not hide additional violations — diagnostic failure samples remain capped at 10, but every non-allow_missing_prior inversion now contributes to causal-ordering totals; covered by a regression test with 20 inversions plus 5 valid pairs.
 - [x] Fix stale storyline source PID handling so skipped create_remote_thread/process_access evidence is reflected in ground truth instead of silently claiming generated evidence.
 - [x] Fix fallback DNS SERVFAIL Zeek packet accounting so rewritten UDP history, packet counts, and IP-byte totals remain consistent.
+- [x] Restore bounded Windows emitter flushing by spooling deferred Windows event dictionaries to an on-disk chronological spool at buffer/barrier flush boundaries, keeping in-memory retention bounded while preserving final RecordID ordering.
 
 ### Skill Installer Agent Support
 
--- a/src/evidenceforge/generation/emitters/windows.py
+++ b/src/evidenceforge/generation/emitters/windows.py
@@ -27,13 +27,13 @@ EventRecordIDs in sorted order (ensuring monotonic IDs match chronological
 order), then renders to XML and writes to per-host FQDN directories.
 """
 
+import json
 import logging
 import os
-import pickle
 import random
 import sqlite3
 import tempfile
-from datetime import datetime, timedelta
+from datetime import UTC, datetime, timedelta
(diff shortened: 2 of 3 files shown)


DavidJBianco (commented):
## Review: Timestamp fixups skipped when spooling (correctness regression)

The memory-bounding motivation is correct and the SQLite-backed spool approach is reasonable. However, `_flush_unlocked` skips the two timestamp fixup passes whenever any events were spooled:

```python
if self._spooled_count:
    self._spool_event_dicts_unlocked()
    events = self._iter_spooled_events_unlocked()
    # ← _shift_process_terminations_after_dependents() never called
    # ← _shift_logoffs_after_dependents() never called
else:
    self._shift_process_terminations_after_dependents()
    self._shift_logoffs_after_dependents()
    events = sorted(self._event_dicts, ...)
```

These fixups enforce two invariants in emitted telemetry:
- Child process-create timestamps come after their parent's process-create
- Logoff events come after the session's last dependent event

Both are visible correctness properties in the output. Since spooling triggers as soon as `_event_dicts` exceeds `buffer_size`, **any scenario large enough to need memory bounding will also skip these fixups** — i.e., the regression affects the common case.

### Fix

The fixups need to run on the full sorted event sequence regardless of spooling. One approach: drain the spool into a list, apply both fixup passes, then write the fixed events back:

```python
if self._spooled_count:
    self._spool_event_dicts_unlocked()
    event_list = list(self._iter_spooled_events_unlocked())
    # run fixups on the in-memory list
    self._shift_process_terminations(event_list)
    self._shift_logoffs(event_list)
    events = iter(event_list)
else:
    self._shift_process_terminations_after_dependents()
    self._shift_logoffs_after_dependents()
    events = sorted(self._event_dicts, ...)
```

The fixup methods would need to accept a list argument rather than operating on `self._event_dicts` directly — a minor refactor, but it keeps the correctness invariants intact for both paths.

### Secondary concern: `pickle` serialization

The spool uses `pickle.dumps/loads` for serialization. Since the spool contents derive from scenario YAML, JSON would be more appropriate and auditable here — the event dicts contain only primitive types (strings, numbers, datetimes) that serialize cleanly.

PR #122, opened 2026-05-11
fix: bound cross-source eval joins

DavidJBianco:
### Motivation
- The cross-source field-agreement scorer performed exhaustive A×B joins for each pivot key, allowing an attacker-controlled eval input to create a high-cardinality pivot bucket and force quadratic CPU work. 
- The change prevents eval-time CPU exhaustion while preserving the intended agreement logic and diagnosable failure reporting.

### Description
- Add configuration-backed caps: `_MAX_PIVOT_BUCKET_RECORDS` limits B-side records retained per pivot key and `_MAX_FIELD_AGREEMENT_MATCHES` limits total pair comparisons per configured pair. 
- Truncate B-side buckets in `_build_pivot_index(...)` to at most `max_bucket_records` and skip indexing beyond the cap. 
- Stop pairwise comparisons early in `_score_pair(...)` once the `max_matches` limit is reached to bound A×B work. 
- Add regression tests in `tests/unit/test_eval_cross_source_pairs.py` that verify pivot-bucket truncation and the global match cap behavior, and mark the TODO tracking entry as resolved in `TODO.md`.

### Testing
- Ran the focused unit tests with `uv run pytest --no-cov tests/unit/test_eval_cross_source_pairs.py` and all tests passed. 
- Executed `uv run ruff check .` and `uv run ruff format --check .` and both checks passed. 
- Running the targeted pytest file with coverage produced a repository-wide coverage failure when the full coverage threshold was enforced because only the one test file was executed, but the behavioral tests added here succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a01ebb4662483258b91de2962618002)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -215,6 +215,7 @@ Verification is complete: dedicated `tests/unit/test_world_model.py` coverage wa
 - [x] Security: bound baseline failed-logon synthetic service account selection loops to prevent scenario-controlled infinite loops/DoS.
 - [x] Security: guard persona `activity_intensity` normalization against all-zero values to prevent divide-by-zero DoS during generation (all-zero overrides now safely map to floor probability instead of crashing).
 - [x] Security: prevent nmap CIDR fallback host expansion from materializing full `network.hosts()` lists (large attacker-controlled CIDRs can trigger memory/CPU DoS).
+- [x] Security: bound cross-source field-agreement joins in evaluation to prevent attacker-controlled pivot buckets from causing quadratic CPU exhaustion (per-pivot B buckets and total pair comparisons are capped, with regression coverage).
 - [x] **Re-generation appends to existing output** — CLI now checks for existing `data/`, `GROUND_TRUTH.md`, and `ENVIRONMENT.md` before generation. Prompts user to confirm overwrite or abort. `--force` / `-f` flag skips prompt for automation/AI use.
 
 ### Tier 1: Foundational Correctness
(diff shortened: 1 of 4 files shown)


DavidJBianco (commented):
## Review: `b_fields` composite-key branch is not capped

The issue is real — unbounded O(N²) pivot joins when many records share a key — and the fix logic is correct in principle. However, the bucket cap is only applied in the **single `b_field` path**. The `b_fields` composite-key branch still calls `index[tuple(parts)].append(rec)` directly and never reaches the new cap logic:

```python
# b_fields branch (unchanged) — NO CAP APPLIED:
for field in b_fields:
    parts.append(str(rec.get(field, "") or "").casefold())
index[tuple(parts)].append(rec)   # ← unbounded, bypasses cap

# single b_field branch (fixed) — cap applied:
key = str(rec.get(b_field, "") or "").casefold()
bucket = index[key]
if len(bucket) < max_bucket_records:
    bucket.append(rec)
```

Multi-field pivot keys remain unbounded. The existing test only exercises the single-field path, so it doesn't catch this.

### Fix

Assign `key = tuple(parts)` in the `b_fields` branch and fall through to the shared append+cap block:

```python
if b_fields:
    parts = []
    for field in b_fields:
        parts.append(str(rec.get(field, "") or "").casefold())
    key = tuple(parts)
else:
    key = str(rec.get(b_field, "") or "").casefold()

bucket = index[key]
if len(bucket) < max_bucket_records:
    bucket.append(rec)
```

This applies the cap uniformly to both single-field and composite-key pivots. A test covering the `b_fields` path with a large same-key batch would confirm the fix.

PR #135, opened 2026-05-11
fix: preserve exact web_scan counts

DavidJBianco:
### Motivation
- The `web_scan` periodic generator used `_iter_periodic_ticks` (which counts yielded ticks) while per-tick realism logic (random skips and pauses) could `continue` before a request was emitted, causing count-based scans to emit fewer requests than the scenario `count` and producing inconsistent `request_count` in ground truth.

### Description
- Apply skip/pause randomness only for duration/end-time scans by guarding the skip/pause branch with `if count is None:` in `src/evidenceforge/generation/engine/storyline.py` so `count`-based scans remain exact.
- Add a regression test `TestWebScanGeneration::test_count_based_web_scan_emits_exact_request_count` in `tests/unit/test_bulk_events.py` which injects an RNG that would have forced the skip path and verifies both the number of `generate_connection` calls and the returned `request_count` equal the requested `count`.
- Mark the corresponding TODO entry complete in `TODO.md` and run lint/format fixes to keep imports and formatting consistent.

### Testing
- Ran the new unit test: `uv run pytest tests/unit/test_bulk_events.py::TestWebScanGeneration::test_count_based_web_scan_emits_exact_request_count` and it passed.
- Ran the full `tests/unit/test_bulk_events.py` test file and observed `95 passed` across the file.
- Ran project style checks: `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a01f2b2b6448325aec743aab0a94bbe)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -40,6 +40,8 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Skill Installer Agent Support
 
+- [x] Fix count-based `web_scan` generation so random skip/pause realism does not consume count-limited ticks before emitting requests.
+
 - [x] Add dual-agent `eforge install-skills` workflow — default to Claude project installs, add explicit `--agent claude|codex`, keep Claude project/global behavior, add Codex user-level `~/.codex/skills/` installs, reject invalid Codex/global combinations, and cover installer safety/stale-cleanup behavior with tests.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Already fixed on `dev` — commit `36a00a9` addresses the count-based web_scan skip/pause bypass. No action needed.

PR #154, opened 2026-05-13
[codex] Emit syslog as RFC 5424

DavidJBianco:
## Summary
- Switch generated Linux syslog output to RFC 5424 with year-bearing ISO/RFC3339 timestamps, PRI/version/procid fields, and RFC 5424 sorting semantics.
- Keep `eforge eval` compatible with older BSD/RFC3164 and legacy ISO-style syslog through parser-marked legacy fallback paths.
- Tighten strict syslog validation so generated/unmarked syslog must be RFC 5424, and update docs, skill references, fixtures, tests, and TODO status.

## Why
BSD/RFC3164 syslog omits the year. The evaluator previously had to infer that year, which could stamp older generated datasets with the wrong calendar year when evaluated later and distort span-based scoring.

## Validation
- `uv run eforge validate-config`
- focused syslog/eval tests
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest -q` -> `3014 passed, 15 skipped`

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -81,7 +81,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### P1 Syslog BSD Timestamp Year Inference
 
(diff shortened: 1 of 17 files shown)


DavidJBianco:
Clean-base scope correction: I force-with-lease updated this PR branch so it now contains only `0247cc7 feat: emit syslog as rfc5424` relative to `dev`.

Validation rerun after the scope correction:
- `uv run eforge validate-config`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run pytest -q` -> `3005 passed, 15 skipped`

PR #166, opened 2026-05-18
fix(state): bound session/PID block offsets to prevent far‑future DoS

DavidJBianco:
### Motivation
- Prevent attacker-controlled far-future storyline times from forcing linear catch-up loops that materialize per-time-block caches and can exhaust CPU and memory during allocation of Windows LUIDs, Linux logind session IDs, and Linux PIDs.

### Description
- Replace the minute-scale Windows `_logon_luid_block_offset()` catch‑up cache with a closed-form deterministic arithmetic model using a `block_width` + seeded `jitter` to avoid per-minute dictionary population.
- Replace the four‑hour Linux `systemd-logind` catch‑up loop by delegating to a new `_linux_logind_session_block_offset()` closed‑form helper and use that during session ID allocation.
- Replace the Linux PID block offset catch‑up cache with a deterministic `_linux_pid_block_offset()` arithmetic model (bounded by block width + jitter).
- Add regression unit tests asserting that far‑future allocations do not materialize block‑offset caches and update `TODO.md` to mark this security fix complete.

### Testing
- Ran the focused unit tests: `uv run pytest tests/unit/test_state_manager.py -q --no-cov`, which passed (58 passed). 
- Ran lint/format checks: `uv run ruff check .` and `uv run ruff format --check .`, which passed. 
- Ran the complete test suite without coverage: `uv run pytest -q --no-cov`, which completed successfully (2984 passed, 37 skipped). 
- Note: a prior run with coverage (`uv run pytest -q` / coverage) failed due to global coverage thresholds unrelated to these functional changes, so the change was validated with the focused unit tests and full test run without coverage.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0557047fa08325a537ef892921f5c5)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -46,6 +46,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Fix unbounded session offset allocation caches for far-future scenario times — replaced minute/four-hour catch-up caches with direct deterministic offsets and covered far-future allocation cases.
 - [x] Harden Windows spool JSON encoding and flushing against scenario-triggered denial-of-service — replaced collision-prone datetime sentinels with typed spool field wrappers and kept final spooled rendering on SQLite-backed streaming fixup passes instead of materializing all rows.
 - [x] Fix TCP DNS fallback packet overhead so TCP SERVFAIL accounting preserves TCP/IP header distributions.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Reviewed against `dev`. The issue is real and still present there; the fix targets the owning state-management layer and is appropriate to merge. Retargeted to `dev` for merge.

PR #167, opened 2026-05-18
fix(ecar): bound parent-order normalization to avoid infinite timestamp shifts

DavidJBianco:
### Motivation
- Prevent an infinite loop in eCAR finalization where untrusted/raw `PROCESS/CREATE` records that self-reference or form cycles force repeated timestamp increments and hang generation.  
- Ensure emitter-level normalization is resilient to arbitrary `objectID`/`actorID`/`ppid` values supplied by raw storyline events without changing higher-level semantics.  

### Description
- Hardened `_normalize_process_parent_order()` in `src/evidenceforge/generation/emitters/ecar.py` to build an explicit graph of visible `PROCESS/CREATE` records, detect cycles, and treat cyclic edges as unshiftable instead of iterating forever.  
- Bounded the number of parent-ordering passes to at most the number of visible `PROCESS/CREATE` records and skip adjustments for keys flagged as part of a cycle.  
- Added regression tests in `tests/unit/test_ecar_spec_compliance.py` that cover self-parent and mutual-parent cycles to verify `EcarEmitter.close()` completes and preserves original timestamps for cyclic records.  
- Updated `TODO.md` to mark the security fix as completed.  

### Testing
- Ran the eCAR spec unit file with `uv run pytest --no-cov tests/unit/test_ecar_spec_compliance.py` and all tests in that file passed (`33 passed`).  
- Ran the targeted subset with `uv run pytest --no-cov tests/unit/test_ecar_spec_compliance.py -k 'parent_process_create or parent_cycle'` and those tests passed.  
- Noted that running the same targeted tests under the repository-wide coverage gate (`pytest` with coverage enabled) fails the global coverage threshold (not the tests themselves); the fix was validated with `--no-cov` to avoid the global coverage fail-under blocking focused verification.  
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .`, which both succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0398e0d7348325bec2c6b6cb800c00)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -46,6 +46,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Fix cyclic eCAR process parent normalization denial-of-service — added cycle detection and bounded parent-ordering passes, with self-parent and mutual-parent raw eCAR regression coverage.
 - [x] Harden Windows spool JSON encoding and flushing against scenario-triggered denial-of-service — replaced collision-prone datetime sentinels with typed spool field wrappers and kept final spooled rendering on SQLite-backed streaming fixup passes instead of materializing all rows.
 - [x] Fix TCP DNS fallback packet overhead so TCP SERVFAIL accounting preserves TCP/IP header distributions.
 - [x] Fix unbounded Sysmon parent process-create shift loop for cyclic raw ProcessGuid relationships with cycle detection and bounded parent-ordering passes.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing after re-evaluation against current dev. The underlying parent-ordering issue is real, but dev already contains the broader accepted parent-order hardening from the earlier integration; this branch is duplicate/narrower and is no longer needed.

PR #170, opened 2026-05-18
fix: bound site map hex token expansion

DavidJBianco:
### Motivation
- Prevent CPU exhaustion from unbounded `{hex8}`/`{hex16}` token expansion in project-local site map overlays by removing the repeated `while token in path` + `str.replace(..., 1)` pattern. 
- Project-local `.eforge` overlays can reach the site-map loader, so a crafted path with many tokens could force quadratic string-copy work and hang generation jobs. 

### Description
- Added a linear ` _replace_token_occurrences()` helper that splits the path on a token and performs a single pass join to replace every occurrence without repeated full-string scans. 
- Updated ` _replace_hex_tokens()` to call ` _replace_token_occurrences()` for each token type (`{hex8}`, `{hex16}`) instead of looping with `path.replace(..., 1)`. 
- Added unit regression tests in `tests/unit/test_site_maps.py` (`TestHexTokenReplacement`) to validate correctness and bounded behavior for large token counts and stable per-occurrence token generation. 
- Marked the related security TODO entry as resolved in `TODO.md` to reflect the remediation. 

### Testing
- Ran `uv run pytest --no-cov tests/unit/test_site_maps.py -q`, and the suite passed (`20 passed`).
- Ran `uv run ruff check .` and `uv run ruff format --check .`, and both checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bfa73c83258494c44aad771970)


DavidJBianco:

PR #171, opened 2026-05-18
fix: harden observation manifest evaluation

DavidJBianco:
### Motivation

- Prevent attacker-controlled OBSERVATION_MANIFEST.json sidecars from being treated as authoritative during `eforge eval`, which could be used to hide missing traces from causality/coverage scoring. 
- Add conservative validation and containment checks so manifests are only honored when they are clearly bound to the evaluated scenario and dataset root.

### Description

- Restrict manifest discovery and loading in `src/evidenceforge/events/observation_manifest.py` by rejecting symlinked manifests, limiting accepted parents to the output directory or its direct parent, and ignoring oversized files (`MAX_OBSERVATION_MANIFEST_BYTES`).
- Add Pydantic validation for `source_status` counters to ensure known non-negative statuses and fail fast on unexpected values via a `field_validator` on `ObservationManifestEvent.source_status`.
- Bind manifests to scenario metadata by adding `observation_manifest_matches_scenario()` which verifies `scenario_name`, `observation_profile`, `collection_window`, and per-event metadata (index/kind/actor/system/activity/event_types) before accepting a manifest.
- Make `load_observation_manifest()` aware of the evaluated `Scenario` and ignore manifests when the scenario profile is `complete` or when the manifest does not match the scenario.
- Update evaluation wiring to call `load_observation_manifest(self.output_dir, self.scenario)` in `src/evidenceforge/evaluation/engine.py` so manifests are loaded with scenario binding.
- Harden scoring by defensively dropping mismatched in-memory manifests inside `CausalityScorer.score()` so an injected `EvaluationContext` cannot exempt events unless the manifest matches the scenario (`src/evidenceforge/evaluation/pillars/causality.py`).
- Add regression tests covering symlinked manifests, complete-profile ignoring, mismatched manifest metadata, and forged in-memory manifests; and update `TODO.md` to mark the security fix complete.

### Testing

- Ran targeted unit tests: `uv run pytest --no-cov tests/unit/test_observation_manifest.py tests/unit/test_eval_cross_source.py::TestObservationAwareCausality -q`, which passed.
- Ran lint/format checks: `uv run ruff check .` and `uv run ruff format --check .`, and applied formatting fixes where needed.
- Ran full unit test suite: `uv run pytest --no-cov -q`, which completed successfully (all unit tests passed: `3167 passed, 37 skipped`).
- Changes are covered by new unit tests added to `tests/unit/test_observation_manifest.py` and `tests/unit/test_eval_cross_source.py` and by existing causality tests.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2d260c83259deb2fecc42679a1)


DavidJBianco:

PR #174, opened 2026-05-18
fix: bound Linux PID block offsets

DavidJBianco:
### Motivation
- Prevent unbounded CPU/memory work during Linux PID allocation when storyline events include far-future absolute timestamps that produce huge elapsed-second blocks.
- Preserve realistic, deterministic Linux PID behavior while removing the attack surface where an attacker-controlled timestamp could force one cache entry per 5-minute block.

### Description
- Replace the previous incremental cache-fill implementation in `_linux_pid_block_offset()` with a bounded direct computation that returns a monotonic block offset formula plus a stable jitter, avoiding materializing intermediate blocks.
- Keep the visible properties of offsets (strictly increasing, small jitter) so allocation semantics and realism are preserved while removing O(block) work and storage.
- Add a regression unit test `test_linux_pid_allocation_does_not_materialize_intermediate_time_blocks` to `tests/unit/test_pid_allocation_realism.py` which allocates a Linux process at year 9998 and asserts the block-offset cache is not populated with intermediate entries.
- Update `TODO.md` to record the security fix as completed.

### Testing
- Ran the focused unit file: `uv run pytest --no-cov tests/unit/test_pid_allocation_realism.py -q`, which passed (`12 passed`).
- Ran the full unit suite: `uv run pytest --no-cov -q`, which passed (`3163 passed, 37 skipped`).
- Ran lint/format checks: `uv run ruff check .` and `uv run ruff format --check .`, which passed.
- All automated tests and checks related to the change succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2d046c8325bb5f781f34e9cbd5)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -50,6 +50,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Fix unbounded Linux PID block allocation for far-future storyline timestamps by replacing materialized intermediate PID block caches with bounded direct offset computation; added a far-future regression test.
 - [x] Harden Windows spool JSON encoding and flushing against scenario-triggered denial-of-service — replaced collision-prone datetime sentinels with typed spool field wrappers and kept final spooled rendering on SQLite-backed streaming fixup passes instead of materializing all rows.
 - [x] Fix TCP DNS fallback packet overhead so TCP SERVFAIL accounting preserves TCP/IP header distributions.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing after re-evaluation against current dev. Current dev already has bounded Linux PID block offset allocation from the prior accepted integration, so this PR is superseded and no longer needs to land separately.

PR #178, opened 2026-05-18
fix: bound ICMP timestamp disambiguation

DavidJBianco:
### Motivation
- Repair a quadratic CPU path in `_disambiguate_icmp_observation_time` where repeated identical ICMP tuple/timestamp collisions were linearly probed by adding 11ms until an unused timestamp was found, which can be driven to O(n^2) work by attacker-controlled high-rate `port_scan` scenarios.

### Description
- Replace the collision-prone `set` probing with a per-rendered-ICMP-tuple `dict` named `_next_icmp_observation_ts_us` and update `_disambiguate_icmp_observation_time` to compute an `adjusted_ts_us` via a single `get()` and advance the per-tuple next-timestamp by 11ms, preserving the existing spacing semantics but making each duplicate assignment constant-time.
- Keep a bounded-map safeguard by clearing the map when its size exceeds 100_000 to avoid unbounded memory growth while preserving prior behavior for normal workloads.
- Add regression coverage `test_disambiguate_icmp_observation_time_uses_constant_time_sequence` to validate constant-time per-observation sequencing and update `TODO.md` to mark the security fix.

### Testing
- Executed the focused regression via `uv run --extra dev pytest --no-cov tests/unit/test_activity.py::test_disambiguate_icmp_observation_time_uses_constant_time_sequence` and the test passed.
- Ran the full activity unit test file with `uv run --extra dev pytest --no-cov tests/unit/test_activity.py` and observed the suite pass (169 tests in that file passed).
- Ran style/format checks with `uv run --extra dev ruff check .` and `uv run --extra dev ruff format --check .`, which both passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2d545c832599b9ec5d3085eb16)


DavidJBianco:

PR #180, opened 2026-05-18
fix: preserve explicit proxy caller PIDs

DavidJBianco:
### Motivation
- A recent change unconditionally invoked the explicit-proxy client helper and could overwrite a caller-provided, valid storyline PID with a User-Agent–derived browser/tool PID, allowing a spoofed UA to misattribute malicious activity to a browser process.
- The intent is to repair known service-owned/stale svchost-style attributions while preserving any legitimate caller/process ownership provided by storylines or earlier generator logic.

### Description
- Change `_caller_explicit_proxy_process_image()` to preserve an existing running process image unless the process is expired, missing, or explicitly identified as service-owned/singleton proxy noise, and return the caller image otherwise.
- Add `_windows_proxy_pid_should_be_replaced(process)` helper to detect known service-owned or singleton Windows service PIDs (svchost-like) and only allow replacement for those cases.
- Adjust `_repair_browser_http_process_attribution()` to short-circuit only when the current PID already matches the expected browser image or when the current PID is *not* a known service-owned proxy PID; otherwise allow the existing caller PID to be preserved.
- Add a unit test `test_browser_proxy_user_agent_preserves_valid_storyline_process` to ensure a storyline-created non-browser process (spoofed by a browser-like UA) remains the initiating PID/image on proxy client legs.
- Update `TODO.md` to mark the security fix as completed.

### Testing
- Ran environment sync and dev deps with `uv sync --extra dev` which installed test/runtime deps successfully.
- Executed the focused tests with `uv run pytest --no-cov tests/unit/test_explicit_proxy.py -q`, which completed with all tests passing (`31 passed`).
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bf03008325b0697037be1c067e)


DavidJBianco:

PR #188, opened 2026-05-18
fix: harden public DNS template rendering

DavidJBianco:
### Motivation
- Prevent local-overlay-controlled public DNS answer templates from causing deterministic crashes (`KeyError`) or large memory allocations via Python `str.format` width/precision fields when generation runs in an untrusted repository.
- Ensure `eforge validate-config` catches unsafe templates before generation so operators/CI do not get a reliable local denial-of-service vector.

### Description
- Add a tiny safe template helper in `src/evidenceforge/config/public_dns_templates.py` that only permits literal text plus the exact `{domain}` and `{domain_hyphen}` placeholders and rejects conversion flags and format specifiers. 
- Wire template validation into the public DNS Pydantic schema so `answer_sets` and `soa_rnames` are validated during `validate-config` using `validate_public_dns_answer_template`. 
- Replace the vulnerable `template.format(...)` sink in `src/evidenceforge/generation/activity/generator.py` with the safe `render_public_dns_answer_template` renderer. 
- Add regression tests to `tests/unit/test_public_dns_profiles.py` and `tests/unit/test_validate_config.py` that assert allowed rendering and that unsafe templates are rejected by both the renderer and `validate-config`, and update the implementation tracker (`TODO.md`).

### Testing
- Ran `uv run pytest --no-cov tests/unit/test_public_dns_profiles.py tests/unit/test_validate_config.py -q` and the targeted tests passed. 
- Ran the full unit/integration test suite with `uv run pytest --no-cov -q`, which completed successfully (all tests passed where expected: `3165 passed, 37 skipped`).
- Ran configuration validation and linters with `uv run eforge validate-config`, `uv run ruff check .`, and `uv run ruff format --check .`, all of which reported success (no validation errors and no ruff formatting issues).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2cdcb0832593a2317113128a13)


DavidJBianco:

PR #189, opened 2026-05-18
fix: guard Zeek drift for non-finite timestamps

DavidJBianco:
### Motivation
- Prevent untrusted raw Zeek `ts` values (`NaN`, `inf`, `-inf`) from causing `int()` conversions that raise and abort multi-sensor generation.

### Description
- Import `math` and short-circuit `_sensor_clock_drift_us()` to return zero when a numeric `ts` is not finite, avoiding `int(float('nan'))` / `int(float('inf'))` paths. 
- Add a focused regression test `test_non_finite_numeric_timestamp_does_not_crash_multi_sensor_emit` that emits `NaN`/`inf`/`-inf` `ts` values to a two-sensor `ZeekEmitter` and verifies all rows are written. 
- Update `TODO.md` to mark the Zeek non-finite timestamp hardening as completed.

### Testing
- Ran `uv sync --extra dev` to install dev/test deps and then `uv run pytest --no-cov tests/unit/test_zeek_multiplex.py -q`, which passed (`21 passed`).
- Ran linter/format checks with `uv run ruff check .` and `uv run ruff format --check .`, which succeeded after formatting the updated test file.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bf7d488325a43abd5ff0c10a5a)


DavidJBianco:

PR #190, opened 2026-05-18
fix: harden syslog system_types validation

DavidJBianco:
### Motivation
- The custom allow-list loop in `eforge validate-config` iterated `entry.get("system_types", [])` directly, which raises `TypeError` when an overlay YAML explicitly sets `system_types: null` before Pydantic schema validation runs. 
- The intent is to let the schema validator report malformed `system_types` values while preventing a pre-validation crash caused by `null` or other non-iterable overlay values.

### Description
- Guarded the syslog check in `src/evidenceforge/cli/validate_config.py` by treating an explicit `null` as absent and skipping the custom iteration when `system_types` is not a `list` (added `system_types = entry.get("system_types") or []` and `if not isinstance(system_types, list): continue`).
- Added two regression tests to `tests/unit/test_validate_config.py` to cover `system_types: null` and scalar `system_types` overlays so the code accepts `null` without crashing and rejects non-list scalars via schema validation.
- Recorded the completed security fix in `TODO.md` to track the change in the implementation plan.

### Testing
- Ran the targeted tests with `uv run pytest --no-cov tests/unit/test_validate_config.py -k 'extra_syslog_system_types or invalid_extra_syslog_system_type'` and the three selected tests passed.
- Ran the full `validate_config` unit suite with `uv run pytest --no-cov tests/unit/test_validate_config.py` and all tests passed (`41 passed`).
- Verified code style with `uv run ruff check .` and `uv run ruff format --check .`, which returned clean results.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2e407083259ef304c7b02813f4)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -50,6 +50,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Harden validate-config syslog system_types validation against null or malformed overlay values — guarded the custom system_types loop so null is treated as absent and scalar malformed values fall through to schema validation; added regression coverage for null and scalar overlay entries.
 - [x] Harden Windows spool JSON encoding and flushing against scenario-triggered denial-of-service — replaced collision-prone datetime sentinels with typed spool field wrappers and kept final spooled rendering on SQLite-backed streaming fixup passes instead of materializing all rows.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing after re-evaluation against current dev. Current dev already validates and guards malformed extra-syslog system_types/weights in the accepted extra-syslog work, so this narrower branch is superseded.

PR #191, opened 2026-05-18
fix: validate explicit_credentials fallback usernames to avoid Pydantic crash

DavidJBianco:
### Motivation

- The generator could construct a fallback `User` from an unconstrained `explicit_credentials.target_username`, which allowed values (e.g., `CORP\\Jane Doe`) that violate the `User.username` pattern and caused Pydantic `ValidationError` at generation time, causing availability failures when processing untrusted scenarios.

### Description

- Add a shared username validation regex `_USER_MODEL_USERNAME_RE` and require it to match before constructing synthetic `User` objects from `target_username` or `system.assigned_user` in `_coerce_windows_explicit_credentials_subject` so malformed account text falls through to the safe Administrator fallback. 
- Preserve existing behavior that returns known users from `self._users_by_username` when present and otherwise falls back to `Administrator` when no safe candidate is available. 
- Add a regression unit test `test_generate_explicit_credentials_ignores_invalid_target_for_subject_fallback` covering a domain-qualified `target_username` containing a space to ensure the coercion path no longer raises. 
- Update `TODO.md` to mark the security remediation and follow the project implementation tracking workflow.

### Testing

- Ran focused unit tests for the explicit-credentials case with `uv run pytest --no-cov tests/unit/test_activity.py -q -k 'explicit_credentials'`, which passed. 
- Ran the full test suite with `uv run pytest --no-cov`, which completed successfully (`3163 passed, 37 skipped`).
- Ran static checks `uv run ruff check .` and `uv run ruff format --check .`, both of which passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2d7e9c832581dd4e3abcba6717)


DavidJBianco:

PR #192, opened 2026-05-18
fix: harden extra syslog service rendering

DavidJBianco:
### Motivation

- The extra-syslog renderer treated scenario-provided `system.services` entries as format strings, so attacker-controlled service names containing `{` or `{missing}` could raise `ValueError`/`KeyError` and abort generation.

### Description

- In `render_extra_syslog_message()` treat the `service` parameter values as literal data and do not run `.format()` on them, while preserving developer-controlled nested templates (e.g., `sudo_command`) that may include `{service}` placeholders. (`src/evidenceforge/generation/activity/extra_syslog.py`)
- Add a regression test that verifies contextual `sudo` rendering still substitutes `service` when it comes from the contextual pool and that malformed scenario service names like `{missing}` or `{` do not crash and are rendered literally. (`tests/unit/test_validate_config.py`)
- Update the persistent work tracker to mark the security hardening as completed. (`TODO.md`)

### Testing

- Ran the focused unit tests: `uv run pytest --no-cov tests/unit/test_validate_config.py -k 'extra_syslog_sudo_templates_render_contextual_services or extra_syslog_treats_contextual_services_as_literals'` and both selected tests passed.
- Ran linters/format checks: `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.
- The change is minimal and preserves existing `sudo_command` nested templating behavior while preventing attacker-controlled scenario fields from being interpreted as format templates.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bf50d08325b11e6f03c5dd0517)


DavidJBianco:

PR #193, opened 2026-05-18
fix: ignore malformed process command URLs to prevent ValueError crash

DavidJBianco:
### Motivation
- A stored process `command_line` containing a malformed or out-of-range HTTP URL (e.g., `:99999`, non-numeric port, or invalid bracketed authority) could be reparsed during connection attribution and cause `urllib.parse.urlsplit` to raise `ValueError`, aborting generation.
- The new process→connection inference path called the unsafe helper without guarding against `ValueError`, introducing a denial-of-service crash for crafted scenario inputs.

### Description
- Hardened `_extract_http_url_from_command()` to skip URL candidates that cause `urlsplit()` to raise `ValueError` instead of propagating the exception.
- Hardened `_http_context_from_process_command()` to catch `ValueError` when accessing parsed attributes (hostname/port) and return `None` for malformed command-line URLs so connection generation will not infer HTTP context from invalid data.
- Added regression tests that cover malformed/out-of-range/non-numeric/bracketed authorities in process command URLs and an attributed-connection test that verifies `generate_connection()` no longer crashes when a stored `curl` command contains a bad port.
- Updated `TODO.md` to record the security fix as completed.
- Files changed: `src/evidenceforge/generation/activity/generator.py`, `tests/unit/test_activity.py`, `TODO.md`.

### Testing
- Ran the focused unit tests `tests/unit/test_activity.py::TestProcessHttpCommandCorrelation` and observed all tests pass (`10 passed`).
- Ran repository linters and format checks: `uv run ruff check .` and `uv run ruff format --check .` passed with no issues.
- The regression verifies that malformed command-line URLs are ignored for HTTP context inference and that an attributed TCP connection generation path no longer raises `ValueError`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13befc3c8325b97157b934b27582)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -50,6 +50,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Harden process command URL parsing so malformed command-line ports cannot crash connection generation — skipped invalid command-line URL authorities/ports during HTTP context inference and covered both direct helper parsing and attributed connection generation.
 - [x] Harden Windows spool JSON encoding and flushing against scenario-triggered denial-of-service — replaced collision-prone datetime sentinels with typed spool field wrappers and kept final spooled rendering on SQLite-backed streaming fixup passes instead of materializing all rows.
 - [x] Fix TCP DNS fallback packet overhead so TCP SERVFAIL accounting preserves TCP/IP header distributions.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing after re-evaluation against current dev. PR #194 is the broader accepted malformed-URL hardening and is now integrated in dev commit 4529265, so this PR is duplicate/narrower.

PR #194, opened 2026-05-18
fix: harden process command URL parsing

DavidJBianco:
### Motivation

- Prevent `eforge generate` from crashing when overlay-provided process command templates contain malformed HTTP(S) URLs (e.g., unterminated IPv6 literals or out-of-range ports). 
- The URL extraction and parsing were calling `urllib.parse.urlsplit` and accessing `parsed.port` without handling `ValueError`, which could abort deterministic generation when an attacker-controlled local `.eforge/config/` overlay supplied a bad template.

### Description

- Defensive-guard the URL extraction helper by updating `src/evidenceforge/generation/activity/generator.py::_extract_http_url_from_command` to skip candidates that raise `ValueError` from `urlsplit` or when `parsed.port` access fails. 
- Add defensive parsing to `src/evidenceforge/generation/activity/generator.py::_http_context_from_process_command` so any `ValueError` from reparsing the selected URL returns `None` instead of propagating an exception. 
- Add unit regression coverage in `tests/unit/test_activity.py` for unterminated IPv6 literals, out-of-range ports, and a mixed command string where a malformed candidate is followed by a valid URL to ensure valid candidates are still recognized. 
- Mark the corresponding item in `TODO.md` as completed to reflect the security hardening.

### Testing

- Ran the targeted unit suite with `uv run pytest --no-cov tests/unit/test_activity.py -q` and observed all tests passed (`171 passed`).
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .` and they succeeded with no formatting or lint failures.
- The new tests verify that malformed HTTP candidates are skipped and that malformed-but-followed-by-valid URLs still produce the expected correlation without causing exceptions.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2e329c83259ba50e1e1f0fe813)


DavidJBianco:

PR #195, opened 2026-05-18
fix: handle empty observation profiles

DavidJBianco:
### Motivation

- Observation profiles that are present as empty mappings (e.g., `profiles: { crash: {} }`) were accepted by config/schema validation and name lookup but treated as falsy at runtime, causing `ObservationPolicy` to raise `ValueError` during generation. 
- This produced a reproducible generation-time denial-of-service when project-local overlays supplied syntactically valid but empty profile entries.

### Description

- Make `observation_profile_names()` only advertise profiles that are mapping-backed and add `observation_profile_exists(name: str) -> bool` to check for a configured mapping in `src/evidenceforge/config/observation_profiles.py`.
- Change `ObservationPolicy.__init__` in `src/evidenceforge/events/observation.py` to treat an explicitly configured empty mapping as a valid profile and only raise when the name is truly absent, preserving the default-visible behavior for intentionally-empty profiles.
- Add regression tests in `tests/unit/test_dispatcher.py` that verify an explicitly configured empty profile yields the default visible policy and that unknown profile names still raise.
- Update `TODO.md` to record the remediation per project workflow.

### Testing

- Ran targeted unit tests: `uv run pytest --no-cov tests/unit/test_dispatcher.py::TestObservationProfiles` and `uv run pytest --no-cov tests/unit/test_validate_config.py::TestValidateConfig::test_validate_config_rejects_invalid_observation_profile_source`, both passed.
- Ran repository-wide tests with `uv run pytest --no-cov` and observed all tests passing (`3164 passed, 37 skipped`).
- Ran style checks `uv run ruff check .` and `uv run ruff format --check .`, both of which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2d1e9483259d540bab47185079)


DavidJBianco:

PR #196, opened 2026-05-18
fix: validate auth noise account names

DavidJBianco:
### Motivation
- The auth-noise `scheduled_stale_credentials.account_base_names` YAML allowed arbitrary non-empty strings but runtime `User.username` enforces `^[a-zA-Z0-9._$-]+$`, causing `eforge generate` to crash with Pydantic validation on malformed overlay values.
- Validate-config used the new AuthNoiseConfig but did not enforce the same username constraints, producing a misleading green validation exit for overlays that later abort generation.

### Description
- Add a compiled regex `_AUTH_NOISE_ACCOUNT_NAME_RE` and a `@field_validator` on `ScheduledStaleCredentialsConfig.account_base_names` to require each entry match the scenario `User.username` pattern, rejecting entries with spaces or slashes.
- Add a unit test `test_validate_config_rejects_invalid_auth_noise_account_names` to ensure `validate_config()` reports an `auth_noise.yaml` error when the overlay contains malformed account names.
- Update `TODO.md` to mark the auth-noise validation task done.

### Testing
- Ran the focused unit suite: `uv run pytest --no-cov tests/unit/test_validate_config.py -q` and the added test passed (40 tests passed locally).
- Ran style checks with `uv run ruff check .` and `uv run ruff format --check .` and fixed import ordering; both checks now pass.
- Changes committed and a PR created titled `fix: validate auth noise account names`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bfbe148325abc995dee996156b)


DavidJBianco:

PR #197, opened 2026-05-18
fix: harden syslog RFC5424 version parsing

DavidJBianco:
### Motivation

- The RFC 5424 syslog parser previously accepted an unbounded decimal `VERSION` via `(?P<version>\d+)` and then converted it with `int()`, which can raise `ValueError` or be abused to cause a denial-of-service crash when fed extremely long digit strings.

### Description

- Constrain the RFC 5424 `VERSION` capture to the spec-supported value `1` by changing the parser regex in `src/evidenceforge/evaluation/parsers/syslog.py` from `(?P<version>\d+)` to `(?P<version>1)` so oversized digit sequences no longer reach `int()` conversion.
- Add a regression test `tests/unit/test_eval_parsers.py::TestSyslogParser::test_long_rfc5424_version_is_parse_error_not_crash` that writes a syslog line with a 5000-digit `VERSION`-like field and asserts the parser returns a parse-error record rather than crashing.
- Update `TODO.md` to mark the security hardening item complete.

### Testing

- Ran the focused syslog parser tests with `uv run pytest --no-cov tests/unit/test_eval_parsers.py::TestSyslogParser -q` and the test collection, including the new regression test, passed.
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .`, both of which passed.
- Ran the full unit test suite with `uv run pytest --no-cov -q`, which completed successfully (all unit tests passed with expected skips).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b13bf06888325bcf3bd3c5ba377c0)


DavidJBianco:

PR #202, opened 2026-05-18
fix: harden web scan ids validation against non-mapping overlays

DavidJBianco:
### Motivation
- Validation previously assumed `ids_ua`, `ids_rate`, and per-path `ids` in `web_scan_presets` were mappings and dereferenced them with `.get()` or membership checks, which raises `AttributeError` when an overlay supplies a non-mapping (e.g., a list). 
- A malformed project-local overlay could therefore crash `eforge validate-config` instead of producing structured validation issues, causing a local denial-of-service on validation. 
- The intent is to fail fast with a clear, structured `Issue` for invalid shapes rather than raise exceptions during validation.

### Description
- Add explicit mapping-type guards in `validate_config()` for `ids_ua`, `ids_rate`, and per-path `ids`, appending an `Issue` when the block is not a `dict` and only calling `_record_ids_rule_identity()` when the value is a mapping. 
- Preserve existing required-field checks when the block is a mapping and continue to validate `threshold` semantics for `ids_rate`. 
- Add two unit tests to `tests/unit/test_validate_config.py` that exercise non-mapping `ids_ua` and per-path `ids` overlays and update `TODO.md` to mark the security-fix tracking entry complete.

### Testing
- Ran the focused unit suite with `uv run pytest --no-cov tests/unit/test_validate_config.py -q`, and all tests passed (`41 passed`).
- Ran lint/format checks with `uv run ruff check .` and `uv run ruff format --check .`, and both checks passed. 
- The new regression tests verify the validator now emits structured `web_scan_presets.yaml` `ERROR` issues for non-mapping `ids_ua` and per-path `ids` instead of raising `AttributeError`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2eab848325962f920323b5a199)


DavidJBianco:

PR #212, opened 2026-05-18
fix: key process termination dedup by start time

DavidJBianco:
### Motivation
- Prevent stale state when Windows PIDs wrap and are reused by making termination de-duplication instance-aware instead of only host+PID aware.
- The prior global `(hostname, pid)` suppression could skip termination events for a distinct later process that reused the same PID, leaving running-process state uncleared.

### Description
- Change `_terminated_process_keys` to store `(hostname, pid, start_time)` (with `start_time` allowed to be `None`) and add `_process_termination_recorded(hostname, pid, start_time)` to check de-duplication by process instance.
- Consult the new helper before skipping termination in foreground finalization (`finalize_foreground_process_lifetimes`), per-connection foreground cleanup, and `generate_process_termination`, and record the terminated instance using the process `start_time` after dispatch.
- Add a regression test `test_process_termination_dedup_allows_reused_windows_pid` that terminates a PID, forces allocator wrap/reuse, creates a new process with the same PID, and verifies the second process still emits termination and is removed from `StateManager`.
- Mark the remediation complete in `TODO.md`.

### Testing
- Ran the focused unit test: `uv run pytest --no-cov tests/unit/test_process_lifetimes.py`, which passed (including the new regression test).
- Ran repository checks: `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.
- Ran the full test suite with `uv run pytest --no-cov`, which completed successfully (all tests passed where run: 3163 passed, 37 skipped in the environment used).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0b0f2e3a208325bd0b9acb783c243a)


DavidJBianco:

PR #221, opened 2026-05-19
fix: harden output target marker reads to prevent symlink exfiltration

DavidJBianco:
### Motivation
- Reading `OUTPUT_TARGET.txt` during evaluation previously followed symlinks and echoed raw marker contents in error messages, which allowed an attacker-supplied generated-output directory to leak local files and oversized payloads to the CLI/CI logs. 

### Description
- Add a small maximum marker size constant `MAX_OUTPUT_TARGET_MARKER_BYTES = 512` and reject marker files larger than this limit in `read_output_target_marker()`. 
- Reject marker symlinks and verify the resolved marker path remains inside the evaluated output-root via `Path.resolve()` + containment check before reading. 
- Sanitize the `normalize_output_target()` error message to avoid echoing the raw (potentially attacker-controlled) marker value. 
- Add focused unit tests in `tests/unit/test_output_targets.py` for non-echoing errors, symlink rejection, and size-limit enforcement, and update `TODO.md` to track the security fix. 

### Testing
- Ran linter: `uv run ruff check src/evidenceforge/output_targets.py tests/unit/test_output_targets.py` which passed. 
- Attempted unit test run: `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_output_targets.py`, but the test run could not complete in this environment due to a missing runtime dependency (`PyYAML`) during test collection; the new tests are unit-level and should pass in CI where dependencies are installed. 
- The change is minimal and preserves existing `write_output_target_marker()` behavior and the public `normalize_output_target()` semantics for valid inputs while closing the symlink/size/echoing disclosure vectors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0bc7b804d48325bd31dd80a9cee4c1)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -60,6 +60,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 ### Security Fixes
 
 - [x] Integrate final approved PR review outcomes into `dev` — accepted and adapted the remaining real fixes (#170, #171, #172, #188, #189, #191, #192, #194, #195, #196, #197, #212), then closed duplicate, superseded, or already-addressed PRs (#167, #168, #174, #190, #193, #213) with rationale.
+- [x] Harden eval output-target marker reads against symlink exfiltration and oversized marker-file abuse.
 - [x] Reworked remaining real-but-not-ready PR fixes #165, #169, #204, and #207 against current `dev` — landed sidecar symlink-safe writes, bounded bash template expansion, raw Zeek OCSP optional-field defaults, and generation-safe extra syslog weights with focused coverage.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing as superseded by the corrected implementation already on dev.

The issue is real, but this PR should not be merged unchanged. Its containment check uses the evaluated path itself as the trust boundary, which can reject the legitimate root-level OUTPUT_TARGET.txt marker when evaluation is pointed at the data/ directory. Current dev already has the corrected version: marker symlinks are rejected, marker size is bounded, invalid target errors do not echo marker contents, and the scenario/output root remains the containment boundary for data/ inputs.

Focused verification on current dev passed: uv run --extra dev pytest --no-cov tests/unit/test_output_targets.py tests/unit/test_cisco_asa_emitter.py, plus Ruff check and Ruff format check.

PR #222, opened 2026-05-19
fix: harden Cisco ASA emitter parsing for malformed raw PRI/severity

DavidJBianco:
### Motivation
- Raw `type: raw` scenario events accept attacker-controlled `fields`, so `pri` and `severity` can be non-integer and previously crashed the Cisco ASA emitter during rendering. 
- Prevent generator termination when an untrusted raw record supplies malformed `pri`/`severity` by falling back to safe defaults instead of raising.

### Description
- Defensively coerce `severity` with `int(...)` and fall back to `6` on `TypeError`/`ValueError`, then compute `pri` via `_pri(severity)` when `pri` is missing or invalid in `src/evidenceforge/generation/emitters/cisco_asa.py`.
- Use the sanitized `severity` value when building the `app_name` header so the rendered header is consistent with the computed severity.
- Added a regression unit test `test_emit_raw_invalid_pri_and_severity_falls_back_without_crash` in `tests/unit/test_cisco_asa_emitter.py` that emits a malformed raw ASA record and asserts the emitter renders using default severity rather than crashing.
- Updated `TODO.md` per repository implementation-state rules to record the security fix completion.

### Testing
- Ran style checks with `uv run ruff check src/evidenceforge/generation/emitters/cisco_asa.py tests/unit/test_cisco_asa_emitter.py` which passed. 
- Ran formatter checks with `uv run ruff format --check src/evidenceforge/generation/emitters/cisco_asa.py tests/unit/test_cisco_asa_emitter.py` (file was reformatted and then checks passed).
- Attempted to run the targeted unit test via `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_cisco_asa_emitter.py`, but the local container test bootstrap failed due to missing test-environment dependencies (`ModuleNotFoundError` for `yaml` / import path); the new regression test is included and will run in CI where the full test environment is available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a0c52c367e08325b889dd3683421cf6)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -59,6 +59,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ### Security Fixes
 
+- [x] Harden Cisco ASA raw RFC3164 rendering against invalid `pri`/`severity` values so malformed raw fields cannot crash generation and added regression coverage for malformed raw inputs.
 - [x] Integrate final approved PR review outcomes into `dev` — accepted and adapted the remaining real fixes (#170, #171, #172, #188, #189, #191, #192, #194, #195, #196, #197, #212), then closed duplicate, superseded, or already-addressed PRs (#167, #168, #174, #190, #193, #213) with rationale.
 - [x] Reworked remaining real-but-not-ready PR fixes #165, #169, #204, and #207 against current `dev` — landed sidecar symlink-safe writes, bounded bash template expansion, raw Zeek OCSP optional-field defaults, and generation-safe extra syslog weights with focused coverage.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing as superseded by the stronger implementation already on dev.

The issue is real: malformed raw Cisco ASA pri/severity fields could crash rendering. Current dev already fixes this at the ASA rendering layer using the shared bounded syslog integer coercion helper, so invalid values fall back safely and out-of-range severities stay within the source-native 0-7 range. It also includes focused regression coverage for malformed and out-of-range raw inputs.

Focused verification on current dev passed: uv run --extra dev pytest --no-cov tests/unit/test_output_targets.py tests/unit/test_cisco_asa_emitter.py, plus Ruff check and Ruff format check.

PR #229, opened 2026-05-22
fix: avoid quadratic storyline future-spec slicing

DavidJBianco:
### Motivation
- Storyline execution created `events[i + 1:]` on every loop iteration which copies all remaining events and causes O(N^2) work for long attacker-controlled event lists.
- The copied `future_specs` is only iterated by a narrow Linux termination heuristic, so creating full list slices is unnecessary and is a denial-of-service risk.

### Description
- Replace eager per-iteration list slices with lazy iterators by using `itertools.islice(...)` for `future_specs` in the normal storyline, interleaved single-storyline, and red-herring execution paths in `src/evidenceforge/generation/engine/storyline.py`.
- Broaden the helper signature from `Sequence[Any]` to `Iterable[Any]` for `StorylineMixin._process_has_following_same_host_connection` since it only performs a forward scan.
- Add `import itertools` and update the unit test `tests/unit/test_storyline_command_networks.py` with a regression assertion that the helper accepts iterator inputs.
- Update `TODO.md` to record the security fix tracking entry.

### Testing
- Ran `uv run ruff check .` which completed successfully. 
- Ran `uv run ruff format --check .` which completed successfully. 
- Attempted targeted `pytest` for the modified unit tests but the execution environment lacked a runtime dependency (`yaml`), preventing the test run here; the new regression assertion was added and should pass under a proper test environment/CI run.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a104f7ffbf08325b14b10f22f7e9b15)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix removes quadratic storyline tail copies with iterator-based future-spec scanning. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

DavidJBianco:
Merge-conflict note: while updating this branch against current dev, I kept the iterator-based future-spec scanning from this PR and the already-merged storyline sqlcmd collision fix. Ruff also removed an import that became unused after the merge resolution.

PR #230, opened 2026-05-22
fix: preserve SOF-ELK syslog route ordering during sudo normalization

DavidJBianco:
### Motivation
- A new sudo-session normalizer was being applied unconditionally to all syslog targets and always re-sorted lines with an RFC5424-only sort key, which can reassign RFC3164/SOF-ELK year-partitioned lines to the wrong route and corrupt output chronology.

### Description
- Avoid re-sorting host buffers when the sudo normalizer finds no RFC5424 sudo rows by returning the original line sequence early in `_normalize_sudo_session_lifecycles_for_lines()`.
- Add a regression unit test `test_normalize_sudo_session_lifecycles_preserves_non_rfc5424_order()` to ensure non-RFC5424 (RFC3164/SOF-ELK) lines retain their original ordering, and mark the TODO entry as resolved.

### Testing
- `uv run ruff check .` and `uv run ruff format --check .` both passed in the branch.
- Attempted `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_syslog_family_renderer.py`, but the test run could not complete in this environment due to a missing runtime dependency (`yaml`); the added unit test is present and will run in CI or a fully provisioned development environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10572d60508325a071e5d0a5b9914e)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix preserves SOF-ELK syslog ordering by limiting sudo normalization sorting to RFC5424 rows. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

DavidJBianco:
Merge-conflict note: while updating this branch against current dev, I preserved both accepted syslog regression tests: the SOF-ELK ordering guard from this PR and the already-merged oversized sshd PID guard. No intended behavior change beyond resolving the overlap.

PR #232, opened 2026-05-22
fix(ecar): anchor PROCESS/TERMINATE timing to process start to avoid lifetime inflation

DavidJBianco:
### Motivation
- A recent change caused eCAR `PROCESS/TERMINATE` render times to recompute the process-create timestamp from the termination event, which sampled source timing relative to the terminate canonical time and inflated rendered lifetimes.  
- This resulted in terminate rows being emitted roughly one extra lifetime late, breaking cross-source temporal integrity and causing contradictions with ground-truth process start/end times.

### Description
- Change `_process_terminate_timestamp()` in `EcarEmitter` to derive the rendered process-create timestamp from a synthetic create-anchor `SecurityEvent` at `proc.start_time` instead of using the terminate event as the base.  
- Preserve the existing `canonical_lifetime` logic and compute the terminate `not_before` bound as `max(event.timestamp, process_create_ts + canonical_lifetime)` so visible lifetime is preserved without doubling.  
- Update the targeted unit test (`tests/unit/test_source_timing.py`) to compute `process_time` from an explicit create event and add an upper-bound assertion to guard against doubled-lifetime regressions.  
- Mark the TODO tracking entry as complete and note the fix in `TODO.md` per repository workflow.

### Testing
- Ran linter and formatter checks which passed using `uv run ruff check .` and `uv run ruff format --check .`.  
- Ran the focused unit test selection locally but full `pytest` execution was blocked in this environment due to a missing runtime dependency (`PyYAML` / `yaml`) so the test run could not be completed here.  
- The modified unit test targets `test_ecar_process_terminate_preserves_rendered_lifetime` and `test_ecar_dependent_timestamp_follows_process_create` in `tests/unit/test_source_timing.py` to guard the behavior change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f95788325be0a7511b20c9671)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix anchors eCAR terminate timing to the rendered process-create anchor. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #234, opened 2026-05-22
fix: guard Windows PID normalization for raw Security events

DavidJBianco:
### Motivation
- Raw `raw` storyline events can route attacker-controlled Windows Security fields (EventID 4688 / `NewProcessId`) into emitter fixups, and unbounded `int(text)` on all-decimal PID strings can raise or consume excessive resources and abort generation. 
- Harden the PID normalization path so maliciously large decimal strings cannot crash or destabilize the Windows emitter during flush/fixup passes.

### Description
- Hardened `_windows_pid_hex()` in `src/evidenceforge/generation/emitters/windows.py` to bound decimal-string conversions to a safe length and to catch `ValueError`, returning the original text on failure instead of raising. 
- Added a focused regression unit test `test_windows_pid_hex_rejects_oversized_decimal_pid` in `tests/unit/test_emitters.py` to ensure oversized decimal PID strings do not raise and remain unchanged through normalization. 
- Updated `TODO.md` to record the vulnerability triage and mark the fix as completed per the project workflow.

### Testing
- Ran static checks: `uv run ruff check` on the modified files and the check passed. 
- Ran format verification: `uv run ruff format --check` on the modified files and the check passed. 
- Attempted to run the targeted pytest for the new unit test (`PYTHONPATH=src uv run pytest --no-cov tests/unit/test_emitters.py -k oversized_decimal_pid`) but the test run could not execute in this environment due to a missing runtime dependency (`PyYAML` / `yaml` import), so the unit test was added but not executed here; CI should run the full test suite where the dependency is present.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f84548325a607ee68d7700abd)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix bounds raw Windows PID normalization for malformed Security events. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #235, opened 2026-05-22
fix: preserve actor/source pid ownership for eCAR PROCESS/OPEN normalization

DavidJBianco:
### Motivation
- The Linux eCAR PID-normalization pass was rewriting top-level `pid` using `objectID` before `actorID` for all records, which caused `PROCESS/OPEN` rows to adopt the target process PID rather than the source/actor PID and thus corrupt source attribution in generated telemetry.

### Description
- Change `EcarEmitter._normalize_linux_pid_morphology()` so `rewrite_pid_field()` prefers `actorID` before `objectID` for records where `object == "PROCESS"` and `action == "OPEN"`, preserving actor/source PID ownership while leaving other record types unchanged.
- Add a focused unit test `test_linux_pid_morphology_preserves_process_open_actor_pid` that reproduces the source-vs-target `PROCESS/OPEN` case and verifies the top-level `pid` remains the actor's PID while `properties.target_pid` is rewritten to the target's PID.
- Update `TODO.md` to mark the normalization fix and regression coverage as completed to follow the repository's implementation-state tracking workflow.

### Testing
- Run `uv run ruff check .` which succeeded with no issues reported.
- Run `uv run ruff format --check .` which succeeded and reported files were already formatted.
- Attempted to run the focused pytest invocation `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_ecar_spec_compliance.py -k linux_pid_morphology`, but the test run failed in this environment due to a missing test dependency (`ModuleNotFoundError: No module named 'yaml'`), so the new unit test was added but not executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f8e4c832591c9a6d55fb9bd78)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix uses actor PID for eCAR PROCESS/OPEN source attribution. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #236, opened 2026-05-22
fix: keep unresolved sqlcmd targets as failed attempts

DavidJBianco:
### Motivation
- Unresolved single-label `sqlcmd` targets were assigned a deterministic fallback IP in `10.0.2.x`, but that IP could collide with a real scenario system and turn an intended failed/unrouted attempt into a successful connection to an unrelated host.\
- The change ensures unresolved single-label DB targets remain rendered as failed/unrouted attempts regardless of deterministic fallback IP collisions, preserving dataset and ground-truth integrity.\

### Description
- In `StorylineMixin` (`src/evidenceforge/generation/engine/storyline.py`) added a local `unresolved_single_label_fallback` flag and updated the failure logic so unresolved single-label fallback targets are treated as failed attempts even when the synthesized IP equals a scenario system IP.\
- The connection generation path now computes `failed_private_attempt = unresolved_single_label_fallback or (target_system is None and _is_private_ip(target_ip))`, forcing `firewall` deny context and failure-state rendering for unresolved single-label targets.\
- Added a unit regression test `test_sqlcmd_unresolved_host_collision_still_generates_failed_tcp_attempt` in `tests/unit/test_storyline_command_networks.py` that creates a colliding system on the deterministic fallback IP and asserts the generated connection is still denied (`S0`, `firewall.action == 'deny'`, and `service is None`).\
- Updated `TODO.md` to mark the remediation task complete per project implementation-tracking conventions.\

### Testing
- Ran linter checks with `uv run ruff check src/evidenceforge/generation/engine/storyline.py tests/unit/test_storyline_command_networks.py`, which passed.\
- Ran formatting check with `uv run ruff format --check src/evidenceforge/generation/engine/storyline.py tests/unit/test_storyline_command_networks.py`, which passed.\
- Attempted to run the focused unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_storyline_command_networks.py -k sqlcmd` but the test run could not be executed in this environment due to a missing dependency (`yaml` / PyYAML); the new unit test was added and is expected to pass in CI where dependencies are installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509eb540832599905348ffd88a51)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix keeps unresolved sqlcmd single-label targets as failed attempts even if a fallback IP collides. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #237, opened 2026-05-22
fix: bound internal_url template expansion

DavidJBianco:
### Motivation
- Prevent a denial-of-service in command parameterization where an attacker-controlled `environment.domain` containing `{internal_url}` could cause an unbounded replacement loop during planner-driven command template expansion.
- Ensure planner-created connection-owner processes remain safe when environment-aware parameterization is used for catalog templates that include `{internal_url}`.
- Preserve existing placeholder substitution semantics while guaranteeing termination even for crafted scenario inputs.

### Description
- Replaced the unbounded `while "{internal_url}" in command_line` loop in `ActivityGenerator._parameterize_command_for_system()` with a bounded loop that iterates exactly `command_line.count("{internal_url}")` times and calls `_pick_internal_url_placeholder()` once per original placeholder, preventing self-reintroduction from causing non-termination (`src/evidenceforge/generation/activity/generator.py`).
- Added a unit regression test that sets `activity_gen._ad_domain = "{internal_url}"` and exercises `_parameterize_command_for_system()` to verify replacement completes and returns a substituted URL prefix (`tests/unit/test_activity.py`).
- Updated `TODO.md` to record the investigation and remediation of the `{internal_url}` placeholder expansion issue (`TODO.md`).

### Testing
- Ran static/format checks: `uv run ruff check` and `uv run ruff format --check` on the modified files, both succeeded.
- Added a unit regression test covering the crafted-domain DoS case; it is included in the test suite but `pytest` could not be executed in this environment due to a missing runtime dependency (`yaml`) when attempting to run the focused unit test command (`PYTHONPATH=src uv run pytest --no-cov ...`).
- The change is minimal and preserves existing behavior for normal domains while preventing the non-terminating expansion case; further CI runs with full deps should execute the new unit test.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f837c832586f4e40190f10f7c)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -36,6 +36,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ## Pre-MVP: Consolidated Quality Fixes — IN PROGRESS
 
+- [x] Investigated and remediated `{internal_url}` placeholder expansion denial-of-service in command parameterization by bounding replacements to the original placeholder count and adding a regression test for crafted domain input.
 - [x] Update the global `eforge-assess` Codex skill so iterative loops default to `scenarios/iteration-test`, prefer family-level realism improvements first, and treat automated eval as a regression guardrail rather than an optimization target. Added a progressive-disclosure family iteration reference, refreshed skill metadata, and verified with the lightweight skill validator only.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing this after retargeting to dev and re-evaluating against the rewritten codebase. The internal_url recursion issue is real, but PR #238 is the stronger root-cause fix because it both bounds placeholder expansion and rejects placeholder-tainted domains; this PR only bounds the loop and leaves tainted placeholders in generated output.

PR #238, opened 2026-05-22
fix: bound internal_url placeholder expansion to prevent recursive DoS

DavidJBianco:
### Motivation
- A template parameterization path allowed unbounded replacement of the `{internal_url}` token which could recurse when `environment.domain` contains placeholder text and cause CPU/memory exhaustion. 
- The AD domain used to build replacement URLs was accepted as an unconstrained string, allowing attacker-controlled values to reintroduce the same token into generated replacements.

### Description
- Replace the unbounded `while` loop in `ActivityGenerator._parameterize_command_for_system` with a bounded pass that runs exactly once per original `{internal_url}` token to prevent infinite growth. 
- Harden `_pick_internal_url_placeholder` so that empty or placeholder-tainted AD domains containing `{` or `}` fall back to `corp.local` before constructing replacement URLs. 
- Add a focused unit test `test_parameterize_command_internal_url_placeholder_is_bounded` that sets `_ad_domain` to `"{internal_url}"` and asserts replacements terminate and the resulting command contains two resolved URLs. 
- Record the security task completion in `TODO.md` following the repository workflow.

### Testing
- Ran linting and formatting checks with `uv run ruff check .` and `uv run ruff format --check .`, both of which succeeded. 
- Added a unit test covering the bounded replacement; a local `pytest` invocation was attempted but collection failed in this environment due to a missing runtime dependency (`PyYAML`), so tests could not be executed here. 
- The change is minimal and targeted to the placeholder expansion path and the added test prevents regression once tests are run in an environment with test dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f0db08325a5a1dfbb7a18977c)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix bounds and sanitizes internal_url placeholder expansion at the URL selection and parameterization layer. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #239, opened 2026-05-22
fix: bound Linux PID allocator retries to prevent unbounded CPU spin

DavidJBianco:
### Motivation
- A parent-ordering constraint introduced a rejection condition that, combined with PID normalization near Linux `pid_max`, could cause `_allocate_linux_pid` to loop indefinitely and consume CPU when allocating child PIDs for parents allocated close to `pid_max`.
- The allocator previously had no global retry limit or fail-fast path, making denial-of-service reachable from untrusted scenarios that use large timeline offsets.

### Description
- Add a bounded retry guard (`max_retries = 8192`) to `StateManager._allocate_linux_pid` and raise a clear `StateError` when the retry bound is exceeded to fail fast instead of spinning indefinitely (file: `src/evidenceforge/generation/state_manager.py`).
- Preserve existing allocation heuristics and the `bounded_candidate` fallback; the change only prevents unbounded collision retries while keeping normal allocation behavior intact.
- Update `TODO.md` to track the in-session security hardening task (file: `TODO.md`).

### Testing
- Ran linting with `uv run ruff check src/evidenceforge/generation/state_manager.py TODO.md`, which passed without issues.
- Attempted to run the targeted unit tests with `uv run pytest --no-cov tests/unit/test_state_manager.py -q`, but the test run could not start in this environment due to a missing runtime dependency (`PyYAML` / `yaml`).
- The change is intentionally minimal and keeps allocator paths and unit-test expectations unchanged except that allocator exhaustion now raises `StateError` instead of hanging; full test runs should be executed in CI or a local environment with dependencies installed to validate behavior under edge-case scenarios.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f5dc88325870cba8fdcba17d9)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix adds bounded Linux PID allocator retries with a clear failure mode. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #240, opened 2026-05-22
fix: persist Sysmon ProcessGuid correlation across flush boundaries

DavidJBianco:
### Motivation

- Address a correlation bug where Sysmon `ProcessGuid` values rewritten during flush-time Event 1 timestamp normalization were not persisted, causing follow-on events emitted in later flushes to recompute and mismatch GUIDs.

### Description

- Persist finalized Event 1 ProcessGuid mappings in the Sysmon emitter by adding `self._final_process_guids` and preferring that cache in `_get_stable_process_guid()` before recomputing from timestamps.
- Write finalized `(hostname, pid) -> ProcessGuid` entries inside `_sync_process_guids_to_event1_times()` when Event 1 GUIDs are adjusted so later batches reuse the normalized GUID.
- Add a focused regression unit test `test_event1_time_shift_persists_process_guid_for_later_batches` to assert that an Event 1 GUID rewrite is reflected by subsequent stable lookups.
- Update `TODO.md` to record validation and remediation of the cross-flush ProcessGuid desynchronization per project workflow.

### Testing

- Ran syntax/compile checks with `python -m py_compile src/evidenceforge/generation/emitters/sysmon.py tests/unit/test_sysmon_emitter.py`, which succeeded.
- Attempted to run the targeted unit tests via `uv run pytest --no-cov tests/unit/test_sysmon_emitter.py -k process_guid`, but the test run in this environment failed due to missing runtime dependencies (`ModuleNotFoundError: No module named 'yaml'`) unrelated to the change; the added test remains as regression coverage for CI where dependencies are available.
- The change is minimal and scoped to the Sysmon emitter behavior and includes a unit test to prevent regressions in flushed GUID persistence.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509ee7308325a151b41d4aef5e95)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix persists final Sysmon ProcessGuid mappings across flush boundaries. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #241, opened 2026-05-22
fix: preserve explicit plain-HTTP conn byte overrides

DavidJBianco:
### Motivation
- A recent change unconditionally overwrote connection `orig_bytes`/`resp_bytes` for successful plain-HTTP flows with values derived from `HttpContext`, discarding explicit scenario-author overrides and weakening dataset integrity.
- The intent is to keep HTTP-derived minimums for realism while preserving caller-provided larger byte-count overrides used to model downloads/exfiltration.

### Description
- In `src/evidenceforge/generation/activity/generator.py` the plain-HTTP branch now computes `derived_orig_bytes, derived_resp_bytes = _http_flow_payload_bytes(event.http)` and sets `event.network.orig_bytes = max(event.network.orig_bytes or 0, derived_orig_bytes)` and `event.network.resp_bytes = max(event.network.resp_bytes or 0, derived_resp_bytes)` instead of unconditionally overwriting the values.
- The unit test `tests/unit/test_activity.py` was updated to assert that explicit larger `orig_bytes`/`resp_bytes` remain intact for plain HTTP flows (`test_generate_connection_plain_http_preserves_explicit_larger_bytes`).
- Implementation tracking was updated in `TODO.md` to record and mark this remediation complete, and changes were committed with the conventional-commit message `fix: preserve explicit plain-http conn byte overrides`.

### Testing
- Ran linter and formatter checks which succeeded: `uv run ruff check .` and `uv run ruff format --check .` (and reformatted the modified file via `uv run ruff format`).
- Attempted targeted unit test run with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_activity.py -k "plain_http_preserves_explicit_larger_bytes"`, but the test run failed in this environment due to missing runtime dependency `PyYAML` (`ModuleNotFoundError: No module named 'yaml'`).
- The change is narrowly scoped and the updated unit test should pass when the repository test dependencies are available and `pytest` is run in a fully provisioned environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509eb0fc8325a4189738d294bd1f)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -36,6 +36,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ## Pre-MVP: Consolidated Quality Fixes — IN PROGRESS
 
+- [x] Validate and remediate plain-HTTP conn.log byte override regression (explicit `orig_bytes`/`resp_bytes` discarded). `generate_connection()` now preserves caller-provided larger `orig_bytes`/`resp_bytes` for successful plain HTTP while still enforcing HTTP-derived minimums, with a unit-test regression asserting explicit override retention.
 - [x] Update the global `eforge-assess` Codex skill so iterative loops default to `scenarios/iteration-test`, prefer family-level realism improvements first, and treat automated eval as a regression guardrail rather than an optimization target. Added a progressive-disclosure family iteration reference, refreshed skill metadata, and verified with the lightweight skill validator only.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing this after retargeting to dev and re-evaluating against the rewritten codebase. I do not recommend this fix: current plain-HTTP byte normalization is intentionally driven by HTTP/file-transfer context so cross-source body semantics stay consistent; preserving arbitrary conn byte overrides would reintroduce source inconsistency instead of fixing the owning canonical context.

PR #242, opened 2026-05-22
fix: bound Windows Kerberos spool reorder memory usage

DavidJBianco:
### Motivation
- A recent Kerberos reordering pass reloaded the entire on-disk Windows event spool into memory before filtering, which defeats the emitter's disk-backed memory bound and enables scenario-driven OOM/DoS.  
- The intent is to preserve the Kerberos 4768/4769 ordering fixup while keeping the spool streaming behavior and bounded memory usage.

### Description
- Added a streaming iterator `def _iter_spooled_kerberos_rows_unlocked(self)` that queries SQLite with JSON extraction to return only EventID 4768/4769 rows.  
- Updated `_shift_spooled_kerberos_tgts_before_service_tickets_unlocked()` to use the new iterator and return early when no Kerberos rows are present, leaving the existing reorder/update logic unchanged.  
- Kept the existing spool encode/decode and update path intact so the fix preserves current behavior and on-disk ordering semantics.  
- Marked the task complete in `TODO.md` per the project implementation-state workflow.

### Testing
- Ran `uv run ruff check src/evidenceforge/generation/emitters/windows.py tests/unit/test_emitters.py` which succeeded.  
- Ran `uv run ruff format --check src/evidenceforge/generation/emitters/windows.py tests/unit/test_emitters.py` which succeeded.  
- Attempted `uv run pytest tests/unit/test_emitters.py --no-cov` but the container is missing runtime dependency `PyYAML` resulting in test execution failure in this environment; the change is minimal and unit tests that exercise the emitter reorder path should pass when run in a properly provisioned dev environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509eb5d08325a839cc8601fe4e1a)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix filters Windows Kerberos spool rows in SQLite before reorder processing. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #243, opened 2026-05-22
fix: guard OCSP request-path bounds against overflow

DavidJBianco:
### Motivation
- Prevent generation from crashing when a project overlay supplies non-finite numeric YAML values (e.g., `.inf`) for OCSP request-path bounds that cause `int()` to raise `OverflowError` at runtime.

### Description
- Hardened the runtime coercion helper by having `_bounded_int()` catch `OverflowError` in addition to `TypeError`/`ValueError` so overlay-provided non-finite values fall back safely and still get clamped by `minimum`/`maximum`.
- Applied the same non-finite defensive handling to `_bounded_probability()` so malformed probability overlays also fallback instead of raising.
- Added a focused unit test `test_ocsp_request_path_ignores_non_finite_bound_overrides` that writes a local overlay with `ocsp.request_path.min_encoded_chars: .inf` and verifies `ocsp_request_path()` returns a valid path.
- Updated `TODO.md` to record and mark the hardening task complete per the project workflow.

### Testing
- Ran `uv run ruff check .` and it succeeded with no issues.
- Ran `uv run ruff format --check .` and it reported files are formatted.
- Attempted `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_dhcp_and_certs.py -k ocsp_request_path` locally but the run failed in this container due to a missing runtime dependency (`yaml` / PyYAML); the new unit test is present and is expected to pass in CI where dependencies are available.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f8d508325891af1e9fda675ee)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix catches overflow in OCSP bounded config coercion. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #244, opened 2026-05-22
fix: prevent external scanner profile weight overflow crashes

DavidJBianco:
### Motivation
- Overlay-controlled scanner profile and per-port weights can be individually finite yet sum to infinity (e.g., two values of `1e308`), which causes `random.choices()` to raise `ValueError` and crash generation when selecting scanner profiles or ports. 
- The change validates and sanitizes aggregate weights at validation-time and runtime so untrusted repository overlays cannot cause availability failures while preserving intended behavior for valid configs.

### Description
- Hardened runtime cleaning in `external_scanner_port_profiles()` to return the built-in defaults when the cumulative profile weight is non-finite and to drop profiles whose per-port cumulative weight is non-finite before any `random.choices()` call. 
- Extended `validate-config` in `src/evidenceforge/cli/validate_config.py` to compute and reject non-finite total `external_scanner_port_profiles` weights and to reject per-profile non-finite cumulative port weights with clear `Issue` messages. 
- Added focused unit tests in `tests/unit/test_network_realism.py` and `tests/unit/test_validate_config.py` that exercise overflowing-profile and overflowing-port-weight cases and updated `TODO.md` to record the fix.

### Testing
- `uv run ruff check .` completed successfully. 
- `uv run ruff format --check .` completed successfully. 
- Attempted to run unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_network_realism.py tests/unit/test_validate_config.py` but the run failed in this environment due to a missing `yaml` (PyYAML) dependency; the new tests are present and should run in CI or a developer environment with project dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509ea01883259bf0fb4cbe7f889e)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix guards external scanner aggregate weights before random weighted selection. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #245, opened 2026-05-22
fix: harden bash workflow config coercion at runtime

DavidJBianco:
### Motivation
- Generation could crash when user/project-local `.eforge` overlays contained non-numeric values for `workflow_model.selection_probability` or workflow `weight`, because the code used `float()` without runtime handling and `eforge generate` does not call the separate validation path prior to loading overlays.
- The change prevents a low-severity local denial-of-service (generation failure) by making generation tolerant of malformed overlay values while preserving intended behavior for valid configs.

### Description
- Add safe coercion helpers `_coerce_probability` and `_coerce_positive_float` to `src/evidenceforge/generation/activity/bash_commands.py` that convert config values to bounded floats with fallbacks on `TypeError`/`ValueError`.
- Use `_coerce_probability` for `workflow_model.selection_probability` and `_coerce_positive_float` for workflow `weight` values so malformed overlay values no longer raise during generation.
- Add two focused unit tests to `tests/unit/test_bash_history_noise.py` that assert an invalid `selection_probability` and an invalid workflow `weight` do not crash `pick_bash_session_commands`.
- Update `TODO.md` to record and mark this remediation task as completed.

### Testing
- Ran linter and formatter checks: `uv run ruff check .` and `uv run ruff format --check .` (both passed in the run environment).
- Added unit tests and attempted to run them with `uv run pytest --no-cov tests/unit/test_bash_history_noise.py`, but the test run in this container failed due to a missing runtime dependency (`PyYAML` / `yaml` import), not a code regression; the new tests are in place and are expected to pass when the test environment has the project dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f87508325a326cd3505077bba)


DavidJBianco:
Reworked after the accept-with-changes review. I kept the malformed bash workflow config hardening, then tightened it so probabilities and workflow weights must coerce to finite bounded values; bad strings, infinities, scalar workflow_model overlays, and huge numeric weights now fall back safely instead of reaching random.choices. Focused validation: uv run pytest --no-cov tests/unit/test_bash_history_noise.py -k "selection_probability or invalid_workflow_weight or scalar_workflow_model".

PR #247, opened 2026-05-22
fix: harden RunMRU overlay command templating against format-string DoS

DavidJBianco:
### Motivation
- Project-local overlays may supply arbitrary strings in `runmru_commands` which were previously passed to `str.format(...)`, allowing brace-containing inputs to raise `KeyError`/`ValueError` or trigger large allocations and crash generation. 

### Description
- Replace unsafe `str.format` usage in `_runmru_command` with a bounded substitution that only replaces `{user}`/`{username}` via `re.sub`, leaving any other brace content literal so malformed or malicious format strings cannot be interpreted. The change is in `src/evidenceforge/generation/activity/edr_pools.py` in `_runmru_command`.
- Add unit tests in `tests/unit/test_edr_pools.py` that assert PowerShell-style braces (e.g., `{ Get-Process }`) are preserved as literals and that width/specifier payloads (e.g., `{user:1000000000}`) are not interpreted during materialization; update `TODO.md` to mark the remediation task complete.

### Testing
- Static checks passed: `uv run ruff check` and `uv run ruff format --check` succeeded for the modified files.
- Unit tests were added for the new behavior and are included in the test suite, but `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_edr_pools.py` could not be executed in this environment due to missing runtime dependency `PyYAML` (module `yaml`) rather than a test failure.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f6d448325ad88ccb1b36ec0a0)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix uses literal RunMRU placeholder replacement instead of format-string templating. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #248, opened 2026-05-22
fix: prevent expired TLS CA intermediates in generated certificate chains

DavidJBianco:
### Motivation
- Prevent the generator from producing certificate chains where a newly-generated leaf is valid at the scenario time but an attached intermediate/authority profile is already expired, which corrupts TLS realism and downstream training/detection quality.
- Preserve historical realism by still allowing historical event timestamps to select historical (now-expired) issuers when appropriate.

### Description
- Added an optional `event_time: datetime | None` parameter to `pick_issuer(...)` and time-filtered candidate issuers by consulting `certificate_authority_profile(...)` so only issuers whose authority-profile `not_valid_before`/`not_valid_after` window covers the event timestamp remain eligible. 
- Updated certificate generation to pass the canonical event timestamp (`event.timestamp`) into `pick_issuer(...)` so leaf issuance and issuer authority validity are evaluated against the same reference time. 
- Added regression unit tests in `tests/unit/test_dhcp_and_certs.py` asserting expired issuers (e.g., Let's Encrypt R3/E1) are excluded for post-expiration timestamps and remain selectable for historical timestamps. 
- Marked the TODO task as completed in `TODO.md` per project workflow.

### Testing
- Ran linter/format checks which succeeded: `uv run ruff check .` (passed) and `uv run ruff format --check .` (passed).
- Attempted to run the focused unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_dhcp_and_certs.py -k issuer`, but the test run failed in this environment due to missing runtime test dependencies (`ModuleNotFoundError` for `yaml` / PyYAML), so the new tests could not be executed here; they are present and should pass in a CI/dev environment with project dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509e63f483259ee201a5e6de0dd2)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix filters TLS issuer CA profiles by scenario event time. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #249, opened 2026-05-22
fix: cap Kerberos transport weights to prevent overflow

DavidJBianco:
### Motivation
- The Kerberos transport profile allowed arbitrarily large integer weights, which can cause `random.choices()` to raise `OverflowError` when overlays supply huge values.  
- Harden validation and runtime selection to prevent untrusted overlays from crashing generation while preserving existing weighted-choice behavior.

### Description
- Add a cap constant `_MAX_TRANSPORT_WEIGHT = 1_000_000` and enforce it in `KerberosTransportProfile` field validation so `udp`/`tcp` weights larger than the cap are rejected in schema validation (`src/evidenceforge/config/schemas.py`).
- Clamp configured weights to the same cap in the runtime picker and wrap `rng.choices(...)` in a `try/except OverflowError` that falls back to returning `"tcp"` to ensure generation does not crash if validation is bypassed (`src/evidenceforge/generation/activity/kerberos_realism.py`).
- Add unit tests that exercise the runtime clamp (`tests/unit/test_kerberos_realism.py`) and schema validation rejection of oversized transport weights (`tests/unit/test_validate_config.py`).
- Update `TODO.md` to mark the hardening task complete.

### Testing
- Ran linters/format checks: `uv run ruff check ...` and `uv run ruff format --check ...`, both succeeded for the modified files.  
- Attempted to run the affected unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_kerberos_realism.py tests/unit/test_validate_config.py`, but the test run could not complete in this environment due to a missing runtime dependency (`ModuleNotFoundError: No module named 'yaml'`), so full pytest verification was not executed here.  
- The added tests are deterministic and should pass in a normal development environment with project dependencies installed (e.g., after `uv sync`/install).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509edf408325aaef7276b7a26a5a)


DavidJBianco:
Reworked after the accept-with-changes review. The fix now uses a single safe positive-weight coercion path for Kerberos transport and nearby weighted Kerberos profile selection, catches TypeError/ValueError/OverflowError, skips non-finite malformed values, and caps valid huge weights before random.choices. Focused validation: uv run pytest --no-cov tests/unit/test_kerberos_realism.py -k "transport_profile or weighted_profile" tests/unit/test_validate_config.py -k "kerberos_transport" plus uv run pytest --no-cov tests/unit/test_kerberos_realism.py -k "weighted_profile".

PR #251, opened 2026-05-22
fix: guard cron overlay commands against non-string values

DavidJBianco:
### Motivation
- A project-local config overlay can supply non-string `cron_commands` values (e.g., boolean/list/dict) which reach cron rendering helpers and trigger `TypeError` in `shlex.quote()` and `str.lower()`, aborting generation.
- Harden the baseline cron path to treat overlay values defensively so malformed overlays cannot crash the generator while preserving normal behavior for valid string commands.

### Description
- Added `_resolve_cron_command(cron_commands, *, is_rhel_like)` in `src/evidenceforge/generation/engine/baseline.py` to validate that `cron_commands` is a dict, select the distro key, require a `str` result, and return a trimmed non-empty string or `None`.
- Replaced the inline cron command selection with a call to `_resolve_cron_command(...)` and bail out early when it returns `None` to avoid passing non-strings into `_cron_shell_command_line()` and `_cron_workload_process()`.
- Added unit tests in `tests/unit/test_phase5_system_traffic.py` that assert malformed truthy non-string overlay values are rejected and that valid distro-specific strings are selected and trimmed correctly.
- Updated `TODO.md` to track the hardening task per project workflow.

### Testing
- Ran `uv run ruff check .` which passed and `uv run ruff format --check .` which passed, ensuring style and import ordering checks are clean.
- Attempted `uv run pytest --no-cov tests/unit/test_phase5_system_traffic.py -q` and `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_phase5_system_traffic.py -q`, but test execution in this environment failed due to a missing runtime dependency (`ModuleNotFoundError: No module named 'yaml'`), so the new unit tests could not be executed here.
- The change is minimal and focused on input validation and unit coverage; once runtime deps are available, the added tests should run and pass.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1052934a048325a7776fbcbe5c2402)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This keeps cron overlay command resolution source-safe by requiring non-empty strings before command generation. Merging as-is.

PR #252, opened 2026-05-22
fix: guard extra_syslog params flattening in validate-config

DavidJBianco:
### Motivation
- Prevent `validate_config()` from raising `TypeError` when an overlay `extra_syslog_messages.yaml` entry contains malformed `params` values (e.g., scalar `1`) that are not iterable, so such overlays are reported as structured schema errors instead of crashing the validator.

### Description
- Guard the `params` flattening logic in `src/evidenceforge/cli/validate_config.py` so it only iterates list-valued `params` entries (`if isinstance(values, list)`) before extracting string values. 
- Add a regression unit test `test_validate_config_reports_extra_syslog_params_type_errors_without_crashing` in `tests/unit/test_validate_config.py` that injects `params: {bad: 1}` and asserts the validator returns a structured `params` schema error instead of crashing. 
- Update `TODO.md` to mark the hardening task as completed per the project workflow.

### Testing
- Ran lint/format checks: `uv run ruff check src/evidenceforge/cli/validate_config.py tests/unit/test_validate_config.py` and `uv run ruff format --check src/evidenceforge/cli/validate_config.py tests/unit/test_validate_config.py`, which passed (formatting auto-applied where needed). 
- Attempted to run the focused unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_validate_config.py -k extra_syslog`, but execution failed in this environment due to a missing external dependency (`PyYAML` / `yaml`), so pytest could not be completed here; the added test is in place and should pass in CI or a dev environment with dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509e90ac8325b02f063c88f4e4af)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -951,6 +951,7 @@ Data works but experienced analysts spot tells. Grouped by format for efficient 
 
 ### Tier 4: Eval Fixes
 
+- [x] Hardened `validate-config` extra_syslog schedule pre-check to avoid crashing on malformed `params` before schema validation; malformed entries now flow to structured schema errors.
 - [x] Harden temporal causal-account exclusion against non-string SubjectUserName/principal values to prevent evaluator exceptions on malformed logs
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This hardens validate-config so malformed extra_syslog params are handled by schema validation instead of crashing during params flattening. Merging as-is.

PR #253, opened 2026-05-22
fix: prevent transient ssh pid allocator from poisoning Windows hosts

DavidJBianco:
### Motivation
- An SSH/syslog transient PID allocator could initialize a host's PID namespace as Linux before the host's real OS was observed, allowing Windows hosts to later receive impossible Linux-style PIDs and corrupt cross-source correlations.
- The validator only warned on Linux-specific events targeting Windows and generation proceeded on warnings, so malicious or malformed scenarios could trigger the allocator poisoning path.

### Description
- Updated `StateManager.allocate_transient_linux_pid` to accept `os_category: str = "linux"` and initialize the per-host PID allocator using the provided OS category rather than unconditionally forcing Linux, preserving Linux behavior when appropriate via `self._allocate_linux_pid` and allocating Windows-style transient PIDs when the host is Windows.
- Threaded `os_category=system.os_category` through transient-PID callsites in `src/evidenceforge/generation/activity/generator.py` and `src/evidenceforge/generation/engine/baseline.py` to ensure callers pass the host OS when requesting transient PIDs.
- Kept Linux allocator behavior unchanged for Linux hosts and reused the Windows PID increment/wrap logic consistent with `create_process()` when the allocator is Windows.
- Updated `TODO.md` to mark the fix as completed.

### Testing
- Ran `uv run ruff check` on the modified files and it passed successfully.
- Ran `uv run ruff format` and applied formatting; `uv run ruff format --check` passed after formatting.
- Attempted targeted unit tests with `uv run pytest --no-cov tests/unit/test_state_manager.py tests/unit/test_world_model.py`, but the run failed in this environment with `ModuleNotFoundError: No module named 'evidenceforge'` due to test import setup, so unit tests could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509e29888325a7ed91b6cfed97c2)


DavidJBianco:
Reworked after the accept-with-changes review. I removed the nonexistent System.os_category usage and kept the boundary Linux-owned: call sites now pass _get_os_category(system.os), and StateManager.allocate_transient_linux_pid rejects non-Linux hosts or an existing Windows PID namespace instead of silently poisoning it. Focused validation: uv run pytest --no-cov tests/unit/test_state_manager.py -k "transient_pid or transient_syslog_pids" and uv run pytest --no-cov tests/unit/test_baseline_canonical.py -k "transient_syslog_pid".

PR #255, opened 2026-05-22
fix: derive proxy action from proxy cache/policy state (don’t use origin status)

DavidJBianco:
### Motivation
- The `x-proxy-action` helper was inferring proxy-side failure actions from origin HTTP status codes (e.g., 403, 5xx) which caused forwarded origin errors to be mislabeled as proxy denials/gateway errors. 
- Proxy actions should reflect proxy policy/cache outcomes, not the origin server response, to avoid corrupting generated proxy evidence and downstream evaluation.

### Description
- Stop using origin `status_code` to decide proxy failure actions in `_proxy_action_for_context`; the function now maps `DENIED`/`AUTH_REQUIRED`/`GATEWAY_ERROR` only from the proxy `cache_result` and otherwise uses request context to return `tunnel`, `ssl-inspect`, or `forward` (changed in `src/evidenceforge/generation/activity/generator.py`).
- Added regression tests that assert origin `403` and `503` combined with `cache_result="MISS"` yield `forward` rather than `deny`/`gateway-error` (tests added to `tests/unit/test_proxy_referrer.py`).
- Recorded the work as completed in `TODO.md` according to the project tracking workflow.

### Testing
- Ran linting checks with `uv run ruff check .` which succeeded. 
- Ran formatting check with `uv run ruff format --check .` which succeeded. 
- Attempted to run the focused unit test file with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_proxy_referrer.py` but the test run failed in this environment due to a missing runtime dependency (`ModuleNotFoundError: No module named 'yaml'`), so full pytest verification could not complete here; the added tests are deterministic and exercise `_proxy_action_for_context` behavior.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f72848325b523c41285ded3c6)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix derives proxy action from proxy policy and cache state rather than origin status. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #256, opened 2026-05-22
fix: prevent malformed absolute HTTP URIs from crashing browser process hinting

DavidJBianco:
### Motivation
- A scenario-controlled absolute HTTP URI can be malformed (e.g., `http://[:::`) and cause `urllib.parse.urlsplit()` to raise `ValueError`, which aborts generation during Windows HTTP process attribution; this change hardens the owning normalization helper to avoid that denial-of-service.

### Description
- Catch `ValueError` in `ActivityGenerator._browser_launch_uri()` when calling `urlsplit()` on absolute `http(s)://...` URIs and fall back to `uri or "/"` to preserve existing behavior without crashing.
- Add a focused unit test `test_browser_http_client_process_hint_handles_malformed_absolute_uri` that exercises `_browser_http_client_process_hint()` with a malformed absolute URI and asserts the hinting path returns safely.
- Update `TODO.md` to mark the browser URI normalization hardening task complete.

### Testing
- Ran `uv run ruff check src/evidenceforge/generation/activity/generator.py tests/unit/test_explicit_proxy.py` and it passed.
- Ran `uv run ruff format --check src/evidenceforge/generation/activity/generator.py tests/unit/test_explicit_proxy.py` and files are formatted.
- Attempted `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_explicit_proxy.py -k "browser_http_client_process_hint_handles_malformed_absolute_uri or connect_target_browser_hint_uses_origin_https_url"`, but the test run could not be executed in this environment due to a missing runtime dependency (`PyYAML`); the added unit test is present and intended to run under a fully provisioned environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509e9904832592295debd72dd454)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix handles malformed absolute HTTP URI parse failures in browser process hinting. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #257, opened 2026-05-22
fix: preserve datetime awareness in RDP source alignment

DavidJBianco:
### Motivation
- The RDP source-alignment helper `WorldPlanner._align_rdp_source_after_future_workstation_session` normalized times to UTC for comparisons but returned a UTC-aware datetime, which could be compared against a caller-supplied naive `datetime`, causing a `TypeError` in generation paths that pass naive datetimes.
- The intent is to align source-side RDP activity after an already-planned workstation session while preserving the caller's datetime awareness semantics (naive stays naive; aware stays timezone-aware).

### Description
- Adjusted `_align_rdp_source_after_future_workstation_session` to compute `aligned_utc` and then return a value that matches the caller `source_process_time` awareness: if `source_process_time.tzinfo is None` return a naive `datetime`, otherwise convert `aligned_utc` back into the caller's timezone with `astimezone`.
- Added a focused regression test `test_align_rdp_source_after_future_session_preserves_naive_time_awareness` that exercises the future-session alignment path with a naive input `datetime` and asserts the returned value remains naive.
- Updated `TODO.md` to mark the investigation/remediation as completed to follow the repository implementation-tracking workflow.

### Testing
- Ran formatting/lint checks: `uv run ruff check .` passed and `uv run ruff format --check .` passed.
- Attempted to run the targeted unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_world_model.py -k 'align_rdp_source_after_future_session_preserves_naive_time_awareness or find_user_session_handles_mixed_timezone_start_times'`, but the test run could not complete in this environment due to missing runtime test dependencies (notably `yaml`), producing an import error; the added unit test is present and intended to pass in a fully provisioned CI/dev environment.
- The change is minimal and preserves existing alignment behavior while eliminating aware-vs-naive comparison crashes for library/direct callers.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509ec9508325ab9bd1283786d90c)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix preserves caller datetime awareness in RDP source alignment. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #258, opened 2026-05-22
fix: preserve scenario dns_query TTL overrides during DNS normalization

DavidJBianco:
### Motivation
- Storyline-provided `dns_query` events documented that `ttl` may be specified per-answer, but `ActivityGenerator._normalize_dns_context_for_resolver` was replacing most caller-provided TTL vectors for successful answers, causing scenario-authored TTLs to be lost.
- This change restores data-integrity for scenario authors who expect explicit TTL control for A/AAAA/CNAME/MX/SRV/PTR/NULL and internal TXT cases.

### Description
- Update `_normalize_dns_context_for_resolver` to early-return and preserve `dns.TTLs` when the caller already supplied a complete TTL vector matching the answer cardinality. 
- Keep existing AA/rcode and internal-vs-external logic intact and only skip normalization for the complete-TTL case. 
- Add/adjust unit coverage in `tests/unit/test_dns_realism.py` to assert that caller-provided TTLs remain unchanged for external TXT and internal A contexts. 
- Record the fix in `TODO.md` to mark the regression as addressed.

### Testing
- Ran linter and formatting checks with `uv run ruff check .` and `uv run ruff format --check .`, which passed. 
- Executed the focused unit test selection via `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_dns_realism.py -k caller_provided_dns_context_ttls_are_preserved`, but the run aborted due to a missing test dependency (`yaml`) in the execution environment before tests could execute; the new test itself asserts TTL preservation. 
- Changes are small and covered by focused unit assertions; further CI execution in the repository environment (with dependencies installed) is expected to pass the updated test.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1050a03b448325b83603385cde9f76)


DavidJBianco:
Reworked after the accept-with-changes review. The fix is now explicit instead of broad: DnsContext carries preserve_ttls, storyline dns_query sets it only when the scenario author supplied ttl, and resolver normalization preserves only that authored TTL vector while keeping ordinary external TXT/cache TTL countdown behavior intact. Focused validation: uv run pytest --no-cov tests/unit/test_dns_realism.py -k "external_txt_context_ttls_use_resolver_cache_countdown or authored_dns_context_ttls_are_preserved" and uv run pytest --no-cov tests/unit/test_storyline_command_networks.py -k "dns_query_event_marks_explicit_ttl_preserved".

PR #259, opened 2026-05-22
fix: preserve causal follow-up ordering after process visibility clamp

DavidJBianco:
### Motivation
- A recent change clamped individual expanded `after` events to occur after a process-create’s visible source time, but each follow-up was clamped independently which could invert the intended semantic ordering of related audit events (e.g., 4738 appearing before 4720). The change restores dataset integrity by ensuring causal follow-ups keep their original relative order after clamping.

### Description
- Enforce monotonic ordering inside `ActivityGenerator._expand_and_emit` by tracking `previous_after_time` and advancing any expanded `after` event timestamp to be strictly later than the prior expanded event when needed, while preserving the existing `_clamp_after_visible_process_create` behavior. (edited: `src/evidenceforge/generation/activity/generator.py`)
- Updated the implementation tracker in `TODO.md` to record that the remediation was completed. (edited: `TODO.md`)

### Testing
- Ran linter/format checks with `uv run ruff check src/evidenceforge/generation/activity/generator.py` and `uv run ruff format --check src/evidenceforge/generation/activity/generator.py`, both of which passed. (success)
- Attempted to run the unit timing-profile tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_timing_profiles.py`, which could not run in this environment due to a missing dependency (`PyYAML` / `yaml`). (blocked)
- Verified repository formatting/linting and updated `TODO.md` per project workflow; no functional behavior other than the monotonic timestamp adjustment was changed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a105294593c8325bebf91fc1b092b00)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix keeps causal follow-up timestamps monotonic after process visibility clamping. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #260, opened 2026-05-22
fix: prevent handshake-only Kerberos TCP flows from being promoted to success

DavidJBianco:
### Motivation
- Prevent synthetic telemetry corruption where TCP partial-handshake Kerberos flows (`S1`, `SH`, `SHR`, etc.) could emit DC audit events and be retroactively promoted to successful Kerberos sessions, changing `conn_state`, packet/byte accounting, and history.
- Preserve intended audit-count reconciliation for legitimate Kerberos sessions while excluding handshake-only and rejected TCP states from audit emission and promotion.

### Description
- Added TCP handshake/reject-state guards to the Kerberos DC-audit emission helper so `_emit_dc_audit_for_kerberos_connection()` returns early for `S0`, `S1`, `SH`, `SHR`, `REJ`, and `OTH` TCP states. (`src/evidenceforge/generation/activity/generator.py`)
- Tightened the pre-emission predicate (`kerberos_has_response`) so handshake/reject states are not treated as response-bearing Kerberos flows. (`src/evidenceforge/generation/activity/generator.py`)
- Added the same handshake/reject-state guard to the Kerberos audit-count promotion path so tuple-counting does not force partial/rejected TCP Kerberos connections to `SF` and inflate `orig_bytes`/`resp_bytes`/`duration`. (`src/evidenceforge/generation/activity/generator.py`)
- Updated `TODO.md` to mark the Kerberos partial-handshake fix as completed per the project’s implementation tracking workflow. (`TODO.md`)

### Testing
- Ran linter and formatting checks with `uv run ruff check .` and `uv run ruff format --check .`, both succeeded.
- Attempted targeted unit runs with `uv run --with ... pytest` for Kerberos-focused tests; initial run failed due to missing test dependencies, then installed test deps via `uv run --with ...` and executed two Kerberos-focused tests, where the originally failing synthetic-audit behavior was reproduced and used to drive iterative fixes (one Kerberos audit-count test passed, the new partial-handshake test initially failed and was used to refine the guards).
- After the code changes the repository passed the lint/format checks; targeted pytest runs were used during development to validate behavior and guided the final predicate/guard changes described above.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509f1cf483258ade00e08b465e52)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix prevents handshake-only Kerberos TCP flows from being promoted to successful audit evidence. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #261, opened 2026-05-22
fix: keep process source-create timing tied to rendered emitters

DavidJBianco:
### Motivation

- A pre-dispatch call recorded preplanned process-create source timestamps (Sysmon/Windows Security/eCAR) before the dispatcher determined which emitters actually rendered the process-create, allowing hypothetical (unrendered) source timestamps to delay/clamp later network evidence.

### Description

- Remove the pre-dispatch `_record_process_source_create_time(...)` call in `ActivityGenerator.generate_process` so source-create timing is recorded only after dispatch and real emitters have run.
- Stop `_record_process_source_create_time(...)` from invoking planning; it now reads and records only already-populated `event.source_timing` entries (i.e., rendered source timestamps) instead of precomputing them.
- Update `tests/unit/test_storyline_command_networks.py` to assert the zero-emitter path produces no `event.source_timing` and no stored process source-create timestamp.
- Mark the Aardvark timing-regression fix complete in `TODO.md` to follow project tracking requirements.

### Testing

- Ran linter/format checks which succeeded: `uv run ruff check .` and `uv run ruff format --check .` (both passed). 
- Adjusted unit test `tests/unit/test_storyline_command_networks.py` to cover the zero-emitter case; the test will validate the fix when run in a full dev environment with test dependencies installed.
- Attempted targeted pytest run `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_storyline_command_networks.py` but it could not complete in this environment due to a missing test dependency (`yaml` / PyYAML).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509e79988325a0a77442cddb267a)

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -36,6 +36,7 @@ Replaced manual per-emitter field coordination with SecurityEvent intermediate r
 
 ## Pre-MVP: Consolidated Quality Fixes — IN PROGRESS
 
+- [x] Aardvark process source-create timing regression fix — removed pre-dispatch process source timing recording so stored source-create timestamps only reflect rendered emitters; added regression coverage for zero-emitter path.
 - [x] Update the global `eforge-assess` Codex skill so iterative loops default to `scenarios/iteration-test`, prefer family-level realism improvements first, and treat automated eval as a regression guardrail rather than an optimization target. Added a progressive-disclosure family iteration reference, refreshed skill metadata, and verified with the lightweight skill validator only.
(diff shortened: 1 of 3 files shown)


DavidJBianco:
Closing this after retargeting to dev and re-evaluating against the rewritten codebase. I do not recommend this fix as-is: removing preplanned process source-create timing breaks existing process-owned timing guarantees, and any remaining issue should be corrected in observation/source-visibility planning rather than by dropping the timing record.

PR #262, opened 2026-05-22
fix: prevent warm-up extra-syslog quota consumption

DavidJBianco:
### Motivation
- Prevent warm-up-generated but suppressed syslog entries from consuming `max_per_host_window` quotas so capped profiles remain available during the real output window.

### Description
- Gate incrementing the per-host extra-syslog counter so it only updates for events with timestamps inside the visible output window (`if limit_key and ts >= self.start_time`), and mark the related `TODO.md` task complete.

### Testing
- Ran lint/format checks: `uv run ruff check ...` and `uv run ruff format --check ...` which passed; unit tests with `pytest` were attempted but the test run failed in this environment due to a missing optional dependency (`PyYAML`), so `pytest` could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a10509eca3c83259aeff82a88c3ebae)


DavidJBianco:
Accepted as-is after retargeting to dev and re-evaluating against the rewritten codebase. This fix keeps warm-up extra_syslog rows from consuming visible-window quota. I aligned only tracker state with current dev where needed to clear merge conflicts; merging as-is.

PR #280, opened 2026-05-27
fix: scope persisted Sysmon ProcessGuid cache by process lifetime

DavidJBianco:
### Motivation
- Finalized Sysmon Event 1 ProcessGuid mappings were previously cached using only `(hostname, pid)`, which allowed a later process that reused the same PID to incorrectly inherit an older lifetime's ProcessGuid and corrupt Sysmon lineage. 
- The intent is to persist rewritten Event 1 GUIDs across flushes while preventing cross-lifetime leakage when Windows PIDs are recycled. 

### Description
- Resolve the effective process lifetime timestamp (`ts`) from the state manager first, then consult the persisted cache keyed by `(hostname, pid, ts)` in `_get_stable_process_guid` by replacing the prior `(hostname, pid)` lookup with `(hostname, pid, ts)`. 
- Persist rewritten Event 1 GUIDs using the lifetime-aware key in `_sync_process_guids_to_event1_times` by storing entries as `(hostname, pid_int, ts) -> new_guid`. 
- Update the `_final_process_guids` annotation to `dict[tuple[str, int, datetime], str]` to reflect the new key shape. 
- Add a regression unit test `test_event1_time_shift_cache_scopes_guid_to_process_lifetime` that verifies reused PIDs in a later lifetime do not inherit a prior lifetime's persisted GUID. 

### Testing
- `uv run ruff check src/evidenceforge/generation/emitters/sysmon.py tests/unit/test_sysmon_emitter.py` completed successfully. 
- `uv run ruff format` was applied and `--check` confirms formatting is consistent. 
- Attempting to run the focused unit tests with `PYTHONPATH=src uv run pytest --no-cov tests/unit/test_sysmon_emitter.py -k process_guid` failed in this environment due to a missing runtime dependency (`yaml`), so the new regression test could not be executed here. 
- The change is minimal and lints/formatting pass; the added regression test exercises the new behavior and should pass in CI with project dependencies installed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a16f7b43d008325892e327f93be2d38)

### DIFF
--- a/src/evidenceforge/generation/emitters/sysmon.py
+++ b/src/evidenceforge/generation/emitters/sysmon.py
@@ -746,16 +746,15 @@ class SysmonEventEmitter(LogEmitter):
         deterministic process-create source offset so Event 1 and all follow-on
         events share the same source-native identifier.
         """
-        cached_guid = self._final_process_guids.get((hostname, pid))
-        if cached_guid is not None:
-            return cached_guid
-
         ts = fallback_timestamp
         sm = getattr(self, "_state_manager", None)
         if sm and pid > 0:
             proc = sm.get_process(hostname, pid)
             if proc is not None and proc.start_time <= fallback_timestamp:
                 ts = proc.start_time
+        cached_guid = self._final_process_guids.get((hostname, pid, ts))
+        if cached_guid is not None:
+            return cached_guid
         rendered_create_time = _SOURCE_TIMING.source_time(
             SecurityEvent(timestamp=ts, event_type="process_create"),
             "source.sysmon_process_create",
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Reviewed against current remote `dev` (`907678c1`). The issue is real: a persisted Sysmon ProcessGuid cache keyed only by `(hostname, pid)` can leak across PID reuse. This patch is not safe to merge as-is because the new cache key is based on the final shifted Event 1 `TimeCreated`, while later `_get_stable_process_guid(host, pid, original_process_start)` lookups are keyed by the original process lifetime. That breaks the existing later-batch contract covered by `test_event1_time_shift_persists_process_guid_for_later_batches` unless the source timing recomputation happens to produce the shifted timestamp.

Recommendation: scope the persisted rewrite by lifetime-derived identity without losing the original-start lookup. A safer shape is to compute the pre-rewrite stable GUID for the process lifetime and cache `old_guid -> rewritten_guid` (or an equivalent lifetime identity), so reused PIDs do not collide while follow-on events from later flushes still resolve to the final shifted Event 1 GUID.

DavidJBianco:
Closing this as superseded by the direct dev fix in 199b6865. The original cache key avoided PID reuse leakage, but keyed the persisted Event 1 rewrite by the shifted timestamp, so later follow-on events resolving from the original process lifetime could miss the rewritten ProcessGuid. The dev fix keys the rewrite by the pre-rewrite lifetime GUID, preserving later-batch lookups without leaking across reused PIDs.

PR #287, opened 2026-05-29
fix: harden external parser target marker reads

DavidJBianco:
### Motivation
- Prevent information disclosure from attacker-controlled dataset directories by avoiding direct reads of `OUTPUT_TARGET.txt` that could follow symlinks and print local secret files to logs. 
- Reuse the existing safe marker reader that enforces symlink, containment, and size checks instead of duplicating ad-hoc logic.

### Description
- Replaced direct `marker.read_text()` + `normalize_output_target()` usage with the safe helper `read_output_target_marker()` in `scripts/external_parser.py` so symlink/containment/size protections are applied centrally. 
- Stopped echoing raw marker contents in error messages so unsupported marker values are not printed to logs. 
- Tightened candidate discovery in `scripts/external_parser.py:_find_output_target_marker` to use `is_symlink()`/`is_file()` and rely on the centralized reader to validate and reject symlinks. 
- Added a unit test `test_external_parser_script_rejects_symlink_marker_without_leaking_contents` and updated invalid-marker assertions in `tests/unit/test_external_parser_script.py` to assert no secret-like contents are leaked.

### Testing
- Ran linter/format checks with `uv run ruff check .` and `uv run ruff format --check .` and they passed. 
- Ran the modified unit tests with `uv run pytest --no-cov tests/unit/test_external_parser_script.py` and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a1933f32a748325ac8bad423ef12eec)

### DIFF
--- a/scripts/external_parser.py
+++ b/scripts/external_parser.py
@@ -63,7 +63,7 @@ from evidenceforge.external_parsers.sof_elk_zeek import (
 from evidenceforge.output_targets import (
     OUTPUT_TARGET_FILENAME,
     OutputTarget,
-    normalize_output_target,
+    read_output_target_marker,
 )
 
 INGEST_STEP_TOTAL = 5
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Reviewed this against `dev` per the automation policy.

The issue is real and was still present on `dev`: `scripts/external_parser.py` read `OUTPUT_TARGET.txt` directly, so a symlinked marker could be followed and an invalid marker value could be echoed into logs. The proposed root-cause approach is correct: route marker reads through the existing `read_output_target_marker()` helper and avoid printing raw marker contents.

I did not retarget and merge this PR directly because the branch is based on the squashed `main` release history; retargeting it to `dev` produces a noisy, divergent compare rather than the narrow two-file fix. I applied the equivalent fix directly to `dev` in commit `9d59a887` (`fix: harden external parser output target marker reads`) and verified it with:

- `uv run pytest --no-cov tests/unit/test_external_parser_script.py`
- `uv run ruff check scripts/external_parser.py tests/unit/test_external_parser_script.py`
- `uv run ruff format --check scripts/external_parser.py tests/unit/test_external_parser_script.py`

Closing this PR as superseded by the `dev` commit.

PR #289, opened 2026-05-31
feat: add spillage event type for synthetic credential leakage

### LINKED ISSUE #283: PR Proposal: Add `spillage` event type for synthetic credential leakage
EvidenceForge currently has no way to emit synthetic credentials/PII into
generated logs. This makes it hard to:

  1. Score a log scrubber / DLP / pre-ingest masker — no labeled positives
     to test against.
  2. Train SOC analysts to recognize credential disclosure in logs.
  3. Validate retroactive log-cleanup processes.

Proposal: a new `spillage` EventSpec that slots into `storyline[].events[]`
alongside the existing types, plus an `eforge eval-scrubber` command that
scores a defender's scrubber against the ground truth.

Example:

```yaml
storyline:
  - id: spill-aws
    actor: terrence.ester
    system: WS-TESTER-01
    activity: "AWS creds leaked via bash history"
    events:
      - type: spillage
        carrier: bash_history
        family: aws_iam     # OR `value: "..."` to provide a literal
```

v1 ships 4 families (`password_generic`, `db_uri`, `aws_iam`, `github_pat` —
the GitGuardian 2026 top categories) and 4 carriers (`bash_history`,
`sysmon_cmdline`, `syslog_msg`, `proxy_header`). Ground truth is written as a
companion `spillage_ground_truth.jsonl` next to `GROUND_TRUTH.md`.

Safety: every emitted value must contain a poison marker (`EvidenceForge`,
`EXAMPLE`, `DO_NOT_USE`) or be on a hard-coded whitelist of vendor-published
canonical fakes (AWS docs `AKIAIOSFODNN7EXAMPLE`, Stripe `sk_test_…`,
jwt.io reference tokens, etc.). All hostnames must be RFC 2606 / RFC 5737.
This check runs at scenario-validate time, so it applies equally to
synthesized and user-supplied literal values.

Scope estimate: ~1100 LOC including tests + docs, single PR. Happy to
split if preferred.

I can submit the PR but if you want to sanity check I think your insights would be good for these 3 questions: 

  1. Is `storyline[].events[]` the right slot, or do you want a sibling
     `spillage:` block?
  2. `eforge eval-scrubber` as a top-level command (matches `generate`,
     `validate`, `eval`) — okay?
  3. Any families I should skip in v1 (e.g., AWS) for policy reasons?

Happy to iterate and if this is out of scope or I missed something and its already included close/reject.

Thanks,

Peter

peterhanily:
Closes #283.

## Summary

This PR adds a new storyline event type, `spillage`, that deliberately places a
**synthetic, provably-fake credential** into a chosen location in the generated
logs — the way a real secret accidentally leaks (a key pasted into a shell, a
token logged by an app, an API key in a request URL). Each spill is recorded as a
labeled positive in a machine-readable ground-truth sidecar.

The purpose is to produce **labeled secret-leakage datasets** for evaluating log
scrubbers, DLP / pre-ingest maskers, and detection rules, and for training
analysts to recognize credential disclosure — without ever writing a real or
usable credential to disk. Generation remains fully deterministic (no model calls
at generation time); values are synthesized from data-driven templates and
emitted through the engine's normal modeled generation paths.

> Note: every example value shown below is synthetic and carries a poison marker
> (`EvidenceForgeFake` / `EXAMPLE`). They are non-functional and safe to display.

## What this adds

A typed event inside the existing `storyline[].events[]` model:

```yaml
events:
  - type: spillage
    surface: shell_history        # where the credential becomes observable
    family: aws_iam               # provide EXACTLY ONE of `family` or `value`
    # value: "EvidenceForgeFake-... DO_NOT_USE"   # a literal alternative
```

- **`surface`** is the semantic exposure surface (never an emitter name), mapped
  to the output format that records it:

  | Surface | Where the credential lands | Output format | Encoding |
  |---|---|---|---|
  | `shell_history` | a shell command in bash history | `bash_history` | shell-quoted |
  | `process_command_line` | a process command line (EDR/endpoint telemetry) | `ecar` | shell-quoted (POSIX) / cmd-quoted (Windows) |
  | `syslog_message` | an application syslog message | `syslog` | control-escaped, single-line |
  | `http_request_url` | the URL/query string of an outbound web request | `web_access` | percent-encoded |
  | `http_referrer` | the `Referer` header of an outbound web request | `web_access` | percent-encoded |

  `shell_history` and `syslog_message` are Linux-modeled; `process_command_line`
  and the two `http_*` surfaces are cross-platform (with OS-appropriate
  rendering). The `http_*` surfaces model a request from the actor's host to a web
  server in the environment (a host with `roles: [web_server]`), recorded in that
  server's access log with the actor's host as the client.

- **`family` vs `value` (mutually exclusive).** Supply a `family` to synthesize a
  fresh value per event from a curated template (so no two spills are identical),
  or a literal `value` (which must pass the same safety checks). Supplying both or
  neither is a validation error.

- **Built-in families** (10), all data-driven: AWS IAM key, GitHub PAT (classic
  and fine-grained), GCP API key, Slack token, Stripe test key, JWT, database
  connection URI, generic bearer token, and a generic password. Families, value
  templates, carrier lines, markers, and the host allowlist are user-customizable
  via the standard config-overlay layer — including adding new families — without
  code changes.

### Correlation scope (v1)

Each surface produces only the correlated system activity that is consistent with
how it is modeled, so a dataset never labels standalone evidence as fully
correlated:

- **`http_request_url` / `http_referrer`** — fully correlated: the credential
  rides a real modeled HTTP/S connection (web access log + network evidence).
- **`process_command_line`** — a live, in-window EDR process record with a
  durable, unique process identity. Its carriers are deliberately **local commands
  only** (`aws configure set`, `git config`, `npm config set`, env-var
  assignments, `cmd /c set`, …), so the record never implies an outbound
  connection it does not model.
- **`shell_history`** — a history-**file** artifact (a command left in
  `~/.bash_history`). It may record network-tool commands (`curl`/`scp`/…) as a
  historical record; v1 emits no correlated in-window process/network telemetry
  for the carrier.
- **`syslog_message`** — a standalone application log line.

## Safety

Because this feature writes credential-shaped strings, the central guarantee is
that **no real or usable credential, and no real host, can be emitted.** Every
value (synthesized or literal) is checked by `check_spillage_safety` against four
guardrails:

1. **Marker / vendor-fake requirement.** The value must contain a poison marker
   (`EvidenceForgeFake`, `EXAMPLE`, `DO_NOT_USE`, `EFORGE_TEST`, …) or be a
   documented vendor-published fake. This is enforced **per credential-shaped
   token**: any substring that looks like a secret — both the modeled family
   shapes and a generic high-entropy detector for unmodeled shapes — must itself
   carry an in-token marker. A marker pasted elsewhere on the line does not vouch
   for a real-looking token next to it.
2. **Host allowlist.** Any host embedded in the value (URL, `user@host`, or bare
   domain) must be an RFC 2606 / 6761 reserved domain or an RFC 5737 / 3849 / 1918
   (or loopback / link-local / private) address. Alternate encodings a real client
   would resolve are normalized before the check — percent-encoding, Unicode label
   separators, obfuscated IPv4 (dotless-decimal/hex/octal), IDN/punycode and
   non-ASCII homoglyph hosts, backslash and scheme-relative URLs, and IPv6 zone
   ids — in bare, userinfo, and URL positions, so a real public host cannot hide
   behind an alternate form.
3. **Family regex.** A value declared for a `family` must match that family's
   pattern.
4. **Single-line / control-free.** The value contains no CR/LF, other line
   separators, or control characters, so a spill cannot split a log line or inject
   a terminal escape.

Literal `value:` spills are checked at `eforge validate` and again at generation;
family spills are checked at generation and by `eforge validate-config`, which
synthesizes and re-checks every family template and rejects an overlay that
weakens the marker / vendor-fake / allowlist policy.

**Ground truth never labels a credential that was not written.** A surface that
*cannot* emit the credential is a hard validation error (e.g. a Linux-only surface
on a Windows host, a missing output format, an `http_*` surface with no
`web_server` host). A surface that *fails to* emit at generation time (e.g. a
filtered request that no sensor observes) is recorded as skipped and excluded from
the sidecar.

This is a strong best-effort contract for synthetic data, not a formal proof.
Documented residuals: a bare integer outside a host position is not treated as a
host; a short or low-entropy secret below the high-entropy threshold can still
pass with only a detached marker; entropy detection is heuristic; and a
hand-crafted `jwt` literal with an artificially short, all-lowercase signature can
be mis-read as a host (real and synthesized JWTs are unaffected). The marker
requirement applies to every accepted value regardless.

## Ground truth and evaluation

- A general machine-readable sidecar, `GROUND_TRUTH.jsonl`, is written next to
  `GROUND_TRUTH.md`. Each line is one record with `schema_version`, `kind`, and a
  stable `record_id`; the `schema_version`/`kind` fields keep the sidecar open to
  future structured record types. In v1 it contains spillage labels only — it is
  not a complete mirror of `GROUND_TRUTH.md`. Spillage records carry:
  `storyline_id`, `time` (the actual emitted log-line time), `actor`, `system`,
  `surface`, `family`, `value`, `value_sha256`, `rendered_value` (the exact
  on-disk byte form), `rendered_sha256`, `expected_sources`, and — for the web
  surfaces — `target_system` (the FQDN of the server whose access log holds the
  line).
- `GROUND_TRUTH.md` (the human report) shows only a redacted preview plus a
  SHA-256; the full value never appears there.
- `eforge eval` recognizes spillage events: its causality pillar reads the sidecar
  and confirms each labeled value actually landed in the logs, matching **per
  record** so each labeled spill in a step must be observed independently in its
  expected source (it does not re-run synthesis). A spillage dataset passes
  acceptance.

## Architecture (code vs data)

The event model, validation, surface handlers, deterministic synthesis, safety
guardrails, and ground-truth writing are **code**
(`src/evidenceforge/generation/spillage.py` and the wiring in the model,
generator, storyline, validation, evaluation, and CLI layers). The credential
family definitions, value templates, carrier lines, poison markers, vendor fakes,
and host allowlist are **data**
(`src/evidenceforge/config/activity/secret_families.yaml`), customizable via the
project's config-overlay mechanism. Adding or modifying a family requires no code
change.

## Scope and limitations

- **No scrubber scoring in this PR.** Evaluation verifies *presence* of the
  labeled values; it does not compute precision/recall against any third-party
  tool. If scrubber scoring is added later it will fold into `eforge eval` rather
  than a new top-level command.
- The `http_*` surfaces send the request **directly** to the target web server
  (bypassing an explicit forward proxy) so the recorded client is always the
  actor's host and the credential cannot be rewritten or stripped in transit — a
  deliberate v1 simplification. A proxy-only surface is intentionally out of
  scope for v1.
- `process_command_line` renders OS-appropriate commands (POSIX shell on Linux,
  cmd/PowerShell on Windows) and is keyed to the `ecar` source on all platforms;
  its carriers are deliberately local so the in-window process record implies no
  network it doesn't model.
- This PR also includes two small supporting fixes surfaced while building the
  feature: making host-context FQDN construction idempotent for already-dotted
  hostnames (so an FQDN-named host's output directory isn't doubled), and wrapping
  validator suggestion text so bracketed tokens render correctly. Both have
  regression tests.

## Before / after

Input:

```yaml
- id: spill-aws
  time: "+20m"
  actor: nina.kapoor
  system: APP-SRV-01
  events:
    - { type: spillage, surface: shell_history, family: aws_iam }
```

Generated `bash_history` line (synthetic value, note the baked-in `EXAMPLE`
marker):

```
export AWS_ACCESS_KEY_ID=AKIABG62LK12DEXAMPLE
```

Corresponding `GROUND_TRUTH.jsonl` record:

```json
{"schema_version":1,"kind":"spillage","record_id":"spill-aws#0",
 "storyline_id":"spill-aws","time":"2024-06-03T13:18:23Z",
 "actor":"nina.kapoor","system":"APP-SRV-01","surface":"shell_history","family":"aws_iam",
 "value":"AKIABG62LK12DEXAMPLE","value_sha256":"cf41a85d8c17…",
 "rendered_value":"AKIABG62LK12DEXAMPLE","rendered_sha256":"cf41a85d8c17…",
 "expected_sources":["bash_history"]}
```

## Testing

- 170 unit and integration tests for spillage, including adversarial safety tests
  for host-smuggling (userinfo/IP, obfuscated-IP encodings, IDN/punycode/homoglyph,
  backslash and scheme-relative URLs), credential-shape evasion, control-character
  rejection, the phantom-label discipline (no label without an emitted line),
  cross-platform behavior, and end-to-end generation + evaluation.
- Full suite: 4113 passed / 40 skipped; line coverage 84.52%.
- Slow-workload suite: 13 passed / 1 skipped.
- `ruff check` and `ruff format --check` clean.

## Try it

```bash
uv run eforge validate tests/fixtures/scenarios/spillage.yaml
uv run eforge generate tests/fixtures/scenarios/spillage.yaml
cat output/GROUND_TRUTH.jsonl                     # the machine-readable answer key
grep -n -i spillage output/GROUND_TRUTH.md        # redacted human report
uv run eforge eval output -s tests/fixtures/scenarios/spillage.yaml
```

## Checklist

- [x] Targets `dev`; no version bump on the feature branch
- [x] Conventional Commit title/message
- [x] Tests added (170 spillage cases, incl. adversarial safety tests); full suite green
- [x] `ruff check` / `ruff format --check` clean
- [x] Docs added/updated: `docs/reference/spillage.md`, scenario reference, scenario
      authoring skill, generate/evaluate/config/validate skills, coverage-test
      prompt, CHANGELOG (Unreleased)
- [x] Determinism preserved (no model calls at generation time)


peterhanily:
@DavidJBianco thanks again for the detailed review on #283. This PR implements all seven points: surface naming with your full example surface set, family/value as mutually exclusive, the code-vs-data/overlay split, strict fake-secret safety validation, the general GROUND_TRUTH.jsonl sidecar, scrubber-eval deferred, and the validator/skill/test updates.

A few things worth flagging:

Secret scanning. The diff trips a "Stripe Test API Secret Key" alert on the vendor_fakes list (and the test that references it). That's expected and intentional vendor_fakes lists published, non-functional vendor documentation examples precisely because real scrubbers recognise them, and GitHub's scanner is one such scrubber. I allowed it to push. If you'd rather the repo not carry any scanner-tripping literal, I'm glad to drop that one Stripe value the marker-based families don't depend on it.

The ground-truth sidecar. You'd offered to build the general GROUND_TRUTH.jsonl layer yourself; I took a pass at it here so spillage had something to plug into. Since it's now a general, schema-versioned artifact that future record kinds will extend, I'd genuinely like your sign-off on the contract the schema_version/kind semantics, the record_id format, and the field set and I'm happy to reshape it to whatever you'd prefer to own long-term.

Scope note. Beyond spillage, the PR carries two small supporting fixes it surfaced: an idempotency guard in _build_host_context (an already-FQDN hostname was getting its domain doubled in the output directory) and a Rich-markup fix so validator suggestions render bracketed tokens like [web_server]. Both have regression tests flagging them since they touch shared code and I know you prefer focused PRs; happy to split either out if you'd rather.

I'll continue to work on the other two PR's but just wanted to get this one to you for review first.

Again great work on this, was testing it out a lot while working on this. Lot of use cases!

DavidJBianco:
Thanks again for this. Before we merge, could you please address a few items?

1. **Fix URL host parsing in the spillage safety guardrail.**

   `_extract_hosts` in `src/evidenceforge/generation/spillage.py` appears to parse URL netlocs with a regex that stops at `/` or whitespace, but not at `?` or `#`. That means something like:

   ```text
   EvidenceForgeFake https://evil.com?x=1
   ```

   can be treated as host `evil.com?x=1` rather than `evil.com`, which can bypass the allowlist check. Please fix this and add regression tests, likely in `tests/unit/test_spillage.py`, for query-string and fragment cases.

2. **Include `GROUND_TRUTH.jsonl` in CLI output handling.**

   The generator writes the new sidecar, but the CLI overwrite/staging/rollback/final-output behavior in `src/evidenceforge/cli/commands.py` still appears to be centered on `GROUND_TRUTH.md` plus the existing outputs. Please make `GROUND_TRUTH.jsonl` participate in overwrite checks, backups, rollback cleanup/restore, and final output listing so stale or missing sidecars cannot occur across repeated runs. `GROUND_TRUTH_JSONL_FILENAME` is already defined in `src/evidenceforge/generation/ground_truth.py`, so that may be useful here.

3. **Tighten spillage eval matching when one storyline step contains multiple spills.**

   The eval path appears to group expected spillage records by `storyline_id`, mainly in `src/evidenceforge/evaluation/engine.py`, and `_spillage_record_matches` in `src/evidenceforge/evaluation/pillars/causality.py` can satisfy that group if any expected value is present. If a single storyline entry contains multiple `spillage` events, finding one should not satisfy all of them. Please make matching per spillage record, or otherwise key it narrowly enough that each expected spill must be observed independently. This may also affect the shape in `src/evidenceforge/evaluation/context.py`. Please add a regression test, likely in `tests/integration/test_spillage_generation.py`.

4. **Add/update docs and skill references for the sidecar, eval behavior, and config.**

   Please make sure the user-facing command/skill docs cover:
   - `GROUND_TRUTH.jsonl` as a generated output, including overwrite behavior. Likely places: `commands/eforge/generate.md` and/or related generate docs.
   - `eforge eval` spillage behavior: it depends on `GROUND_TRUTH.jsonl`; without the sidecar, spillage labels cannot be matched. Likely place: `commands/eforge/evaluate.md`.
   - `src/evidenceforge/config/activity/secret_families.yaml` as overlay-customizable config. Likely places: `commands/eforge/config.md` and the config customization reference docs if applicable.

5. **Clarify the current scope of `GROUND_TRUTH.jsonl`.**

   I originally had in mind eventually making this a machine-readable companion for the whole `GROUND_TRUTH.md`, but I realize my issue comment mostly specified the spillage record shape. I don’t think we need to expand this PR to serialize every existing event type.

   Could you please make the docs/code wording explicit that `GROUND_TRUTH.jsonl` currently contains structured machine-readable records for spillage labels only, and is designed to be extended to other ground-truth record kinds later? That will avoid users assuming it is already a complete JSONL mirror of `GROUND_TRUTH.md`. Likely places: `src/evidenceforge/generation/ground_truth.py` and `docs/reference/spillage.md`.

6. **Clean up a small ground-truth wording inconsistency.**

   Some wording says spillage is labeled in `GROUND_TRUTH.jsonl`, “not `GROUND_TRUTH.md`,” but this PR also adds redacted Markdown summaries. Better phrasing would be: full machine-readable labels live in `GROUND_TRUTH.jsonl`; redacted human-readable summaries live in `GROUND_TRUTH.md`. `docs/reference/spillage.md` and the scenario reference docs are probably the main places to check.

Please also add tests for the three behavioral fixes above.

DavidJBianco:
One more issue I found while testing the PR branch locally, plus a related integration concern.

Some `process_command_line` spillage records that were expected in eCAR did not show up in the final output. From tracing it, the spills are emitted to the eCAR buffer correctly, but the per-host post-flush normalization pipeline drops them in `_deduplicate_semantic_events` after `_normalize_linux_shell_foreground_order` shifts the spill CREATE into the same millisecond slot as an existing baseline PROCESS row with the same host/principal/object/action/PID.

Could you please fix this so `process_command_line` spillage events remain distinct and survive eCAR normalization? My initial read is that these spills should get a stable, unique process identity/PID through the normal process execution path, rather than being shiftable into a dedup collision. Please add a regression test that verifies multiple `process_command_line` spills are present in final eCAR output and matched by the sidecar/eval.

Relatedly, could you take another pass at how the non-HTTP surfaces integrate with the existing action-bundle/canonical-event model?

The `http_*` surfaces look well integrated: they route through `generate_connection` with an `HttpContext`, so the network-connection path can produce correlated web/network evidence. The other surfaces currently look more standalone:

- `shell_history` uses `generate_bash_command(..., emit_process_telemetry=False)`. For carriers that represent actual executed commands, especially `curl`, `aws`, `mysql`, `scp`, etc., that means the bash line appears without the correlated process/network activity we’d normally expect from the shell-command/process/network bundles.
- `process_command_line` should behave like a real process execution with durable session/process identity, and network-capable commands should either produce the expected correlated network evidence or avoid carriers that imply network activity in v1.
- `syslog_message` is probably okay as a standalone app log for v1, but if it intentionally has no correlated service/process evidence, the docs should make that limitation clear.

I don’t think this PR needs to become a full ground-up redesign, but I do want the v1 surfaces to either route through the existing bundle/canonical paths where the carrier implies real execution, or be scoped/documented so we don’t label unrealistic standalone evidence as fully correlated system activity.

peterhanily:
Thanks for the detailed review — I've pushed an update that addresses every item, plus some additional safety/realism hardening surfaced while testing. Mapped to your comments:

1. **URL host parsing** — `_extract_hosts` now terminates the netloc at `?`, `#`, `\`, and whitespace (and handles scheme-relative `//host`), with regression tests for query-string and fragment cases in `tests/unit/test_spillage.py`.
2. **GROUND_TRUTH.jsonl in CLI output handling** — the sidecar now participates in overwrite detection, staging, backup, rollback restore/cleanup, and the final output listing in `cli/commands.py` (via `GROUND_TRUTH_JSONL_FILENAME`). A run that produces no spills removes a stale sidecar from a prior run; regression test in `tests/integration/test_spillage_generation.py`.
3. **Per-record eval matching** — `evaluation/engine.py` now carries each labeled value's expected source, and `_all_spillage_values_traced` in `causality.py` requires every labeled spill in a step to be observed independently (in its own source format, by a distinct trace). Tests cover a multi-spill step, identical-value duplicates, and cross-surface non-credit.
4. **Docs / skills** — `commands/eforge/generate.md` (sidecar as a generated output + overwrite behavior), `evaluate.md` (eval depends on the sidecar), `config.md` (`secret_families.yaml` as overlay config), and `validate.md` (spillage validation errors) are updated.
5. **Sidecar scope** — `ground_truth.py` and `docs/reference/spillage.md` now state explicitly that `GROUND_TRUTH.jsonl` holds spillage labels only in v1 and is designed to be extended to other record kinds — not a complete mirror of `GROUND_TRUTH.md`.
6. **Wording** — the "labeled in jsonl, not md" phrasing is gone; the docs now say full machine-readable labels live in `GROUND_TRUTH.jsonl` and redacted human-readable summaries in `GROUND_TRUTH.md`.
7. **process_command_line eCAR normalization** — these spills are now standalone process-execution records with a durable, unique PID rather than foreground children of the actor's interactive shell, so the post-flush normalization/dedup pass cannot shift them into a collision. A regression test asserts multiple process spills survive in final eCAR under a busy session and are matched by the sidecar/eval.
8. **Non-HTTP surface integration** — taking your "avoid carriers that imply network activity" option: `process_command_line` is a live in-window process record and now uses local commands only (`aws configure set`, `git config`, `npm config set`, env-var assignments, `cmd /c set`, `setx`). Network-tool credential leaks remain represented where they're consistent — `shell_history` (a history-file artifact that legitimately records past `curl`/`scp` commands) and the correlated `http_*` surfaces — and `syslog_message` is documented as a standalone app log. Each surface's correlation scope is now spelled out in `docs/reference/spillage.md`.

**Additional hardening** (same classes, found while testing): the host allowlist also normalizes Unicode label separators, non-ASCII/homoglyph hosts in bare and userinfo positions, IPv6 zone ids, and percent-encoding before checking; the single-line check now rejects all control characters (not just CR/LF); and a regression guard prevents any `process_command_line` carrier from reintroducing a network tool.

**On GROUND_TRUTH.jsonl** — you offered to own this layer; I've kept to the field set and `schema_version`/`kind` contract you specified and would welcome your sign-off (or any reshaping) on the schema, since it's meant to grow other record kinds.

**Scope** — this grew while addressing the review. It's all in service of the items above and the safety guarantee, but if you'd prefer the carrier/OS realism work as a separate follow-up to keep this PR tighter, I'm happy to split it.

**Heads-up** — the diff still trips the secret-scanning alert on the Stripe *test* key in `vendor_fakes` — a documented, non-functional vendor example (which is the point; real scanners should recognize it). I can drop that one value if you'd rather; the marker-based families don't depend on it.

Full suite: 4113 passed / 40 skipped, 84.5% coverage; slow-workload suite 13 passed / 1 skipped; `ruff` clean.

DavidJBianco:
Thanks for the substantial follow-up work here. I re-checked the June 2 review points, and you addressed the main items we asked for: the host-parsing fix, `GROUND_TRUTH.jsonl` CLI handling, per-record spillage eval matching, the sidecar/docs wording cleanup, and the earlier `process_command_line` normalization/dedup problem.

We’re going to go ahead and accept this into `dev`.

There is still one narrow issue we want to clean up on the maintainer side: `process_command_line` spillage is not yet tied back into the actor’s real session/logon context as strongly as we want for EvidenceForge’s canonical auth/session/process model. At this point that remaining gap is more about our internal architectural ownership model than about the feature design itself, and it will be faster and less noisy for us to finish that integration directly on `dev` rather than sending you through another review round for a subtle maintainer-specific fix.

Separately, I’ve decided we do want `GROUND_TRUTH.jsonl` to grow into the machine-readable companion to `GROUND_TRUTH.md` more generally, not just for spillage, but we’ll handle that as a separate follow-up PR rather than broadening the scope of this one after the fact.

Thanks again for the careful iteration on this one.

PR #296, opened 2026-06-04
feat: add adversarial_payload event type for log-pipeline weakness testing

### LINKED ISSUE #284: PR Proposal: Add `adversarial_payload` event type for log-pipeline injection testing
Defender log pipelines (parsers, SIEMs, log shippers, terminals, CSV exports
opened in Excel) can themselves be vulnerable to malicious *content* in logs.
There's no synthetic corpus today that embeds known-bad payloads as field
content propagated consistently across log sources.

Proposal: a new `adversarial_payload` EventSpec that injects a known-malicious
string into an existing log carrier. Mirrors the `spillage` event type
https://github.com/Cisco-Talos/EvidenceForge/issues/283 same shape, same ground-truth pattern, same
safety guardrails.

Example:

```yaml
storyline:
  - id: ansi-in-username
    actor: external-attacker
    system: DMZ-WEB-01
    activity: "ANSI escape sequence in username (terminal hijack)"
    events:
      - type: adversarial_payload
        carrier: syslog_msg
        family: ansi_escape
```

v1 ships 4 families chosen by fresh-CVE relevance in 2025-2026:

  - `ansi_escape`         — Tomcat CVE-2025-55754, ActiveRecord CVE-2025-55193,
                            Splunk SVD-2023-0606
  - `csv_formula`         — HackerOne recurring; `defusedcsv` exists because
                            the problem is unsolved
  - `crlf_log_forging`    — CWE-117, Django CVE-2025-48432, foundational
                            primitive
  - `parser_cve_triggers` — Fluent Bit CVE-2025-12977/-12978/-12970 (Oct 2025),
                            syslog-ng GHSA-7932-4fc6-pvmc

5 carriers: `http_ua`, `http_uri`, `syslog_msg`, `csv_field`, `command_line`.

Corpus is a bundled JSON file at `src/evidenceforge/_data/adversarial_payloads/`,
vendored from PayloadsAllTheThings (MIT, attributed). Every payload contains
the literal `EFORGE_TEST` marker so defenders can allowlist it in their own
SIEM rules.

I am not planning to include these v1: JNDI/Log4Shell (already covered by
`huntresslabs/log4shell-tester`), Unicode/BiDi (low fresh-CVE volume), SSTI
(wrong shape for log pipelines), Jupyter/pandas XSS (no recent CVEs). But if you think it's suitable can adjust the plan

Scope: ~600 LOC, single PR, mirrors the spillage PR architecturally.

I can submit the PR but if you want to sanity check I think your insights would be good for these 3 questions:

  1. New `adversarial_payload` event type (mirrors spillage), or `payload:`
     field on existing event specs? I lean toward new type for review symmetry.
  2. Bundle corpus inline (this proposal), or accept an optional external
     dep on a future `log-payload-corpus` PyPI package?
  3. Marker `EFORGE_TEST` embedded in every payload — okay, or do you want
     it user-configurable?

Thanks,

Peter

peterhanily:
## Summary

Adds a typed `adversarial_payload` storyline event — the counterpart to the `spillage`
event type. Where spillage leaks a provably fake credential, this injects a known
log-pipeline weakness payload into a chosen semantic exposure surface and records it as a
labeled positive in canonical machine-readable ground truth.

The goal is labeled datasets for testing whether parsers, SIEMs, log shippers, terminals,
CSV/spreadsheet exports, SQL-backed stores, and IDS pipelines handle untrusted log content
safely.

Closes #284.

## How it works

- **Deterministic.** Payloads are synthesized from data-driven families in
  `config/activity/payload_families.yaml` and emitted through the normal modeled
  generation paths.
- **Safe by construction.** Every physical line must carry a poison marker
  (`EFORGE_TEST`), embedded hosts default to the inert non-resolving canary
  `canary.eforge.invalid`, and safety is checked at config load, validation, and
  generation.
- **Supported families.** All included families are first-class supported payload
  families: `ansi_escape`, `crlf_log_forging`, `csv_formula`, `log4shell`,
  `xss_reflection`, `sql_injection`, `structured_log_injection`, and
  `oversized_field`.
- **Live callback mode.** `--oob-host` is the explicit opt-in for operator-controlled
  OOB testing. It requires a concrete registrable domain or IP literal, is also
  supported by `eforge validate --oob-host`, and is rejected at the lower safety boundary
  if too broad.
- **On-wire IDS modeling.** Cleartext HTTP payloads may emit a Snort/Suricata alert only
  when a signature-mapped family is observable by an IDS sensor on the path; ground truth
  and `snort_alert.log` stay aligned.

## Surfaces

| Surface | Records in | Notes |
|---|---|---|
| `syslog_message` | `syslog` | Linux-modeled raw injection surface |
| `process_command_line` | `ecar` | Process command-line telemetry |
| `http_user_agent` / `http_request_url` / `http_referrer` | `web_access`, and optionally observed `zeek_http` for cleartext HTTP | Cross-platform; optional `scheme: http | https` |

## Tests & docs

Includes schema validation, config validation, generation, evaluation, ground truth,
reference docs, and skill guidance. The latest review round added regression coverage for
OOB allowlist enforcement at the safety boundary and for `expected_sources` only naming
sources that are actually emitted in the generated dataset.


DavidJBianco:
@peterhanily Thanks for the substantial work here. I reviewed the implementation and I’m inclined to accept the new capabilities in this PR, including the broader family set, IDS integration, and optional OOB/live-callback mode. Before merge, I’d like a few cleanup and hardening items addressed.

1. **Remove the `proposed` mechanism entirely**

Since we’re accepting the added families, please remove the “proposed” concept rather than leaving it as runtime/user-facing uncertainty.

Please remove:
- `proposed` from the payload family schema
- `proposed: true` entries in `payload_families.yaml`
- validation warnings for proposed families
- tests that assert proposed-family warnings/behavior
- docs language like “proposed” / “pending maintainer sign-off”

If a family is included in this PR, it should be treated as supported.

2. **Tighten `--oob-host` validation**

The live callback mode is useful, but the current OOB host validation is too permissive.

Today `--oob-host` checks that the value is a bare host with no scheme/path/userinfo/port, but it does not check that the host is narrow enough. For example, `--oob-host com` is accepted. That value is then added to the safety allowlist, and the inherited host allowlist logic allows either an exact match or a suffix match. As a result, a payload containing `evil.com` can pass because `evil.com` ends with `.com`.

That means a user can accidentally or intentionally turn an OOB registration into a very broad allowlist for an entire TLD/public suffix namespace. This weakens the safety boundary of live callback mode.

Please require a concrete registrable domain such as `example.com` / `oast.fun`, or an IP literal. Do not accept bare TLD or TLD-like entries such as `com` or `co.uk`. Adding a hostname or subdomain such as `myhost.example.com` or `myhost.mysubdomain.example.com` should be acceptable, but not required.

Please add regression tests covering the OOB host contract:
- Reject bare TLD / single-label entries, such as `com`, `fun`, and `local`.
- Reject TLD-like / public-suffix-style entries, such as `co.uk`.
- Accept concrete registrable domains, such as `oast.fun` or a similar test fixture domain.
- Accept concrete hostnames/subdomains, such as `myhost.oast.fun` and `myhost.mysubdomain.oast.fun`.
- Verify suffix matching is scoped only under the accepted concrete domain: `oast.fun` can allow `abc.oast.fun`, but not `evil.com`; `com` must not allow `evil.com`.
- Preserve IP literal support, such as `127.0.0.1`.

3. **Remove `--i-am-authorized`**

I don’t think `--i-am-authorized` adds enough value to keep. It is only a boolean acknowledgement; it does not verify host ownership, target authorization, or anything else. I’d prefer to remove it and make `--oob-host` itself the explicit opt-in.

Please keep the loud `LIVE CALLBACK MODE` warning when `--oob-host` is used, and make the docs clear that OOB mode should only be used against systems the operator is authorized to test.

4. **Add validation parity for OOB literals**

Right now `eforge generate --oob-host ...` passes OOB hosts into `ScenarioValidator`, but standalone `eforge validate` has no equivalent option. That means a scenario with a literal OOB payload can fail standalone validation but pass generation validation.

Please add matching validate support, preferably:

```bash
eforge validate scenario.yaml --oob-host example.com
```

That keeps “validate before generate” reliable for live-callback scenarios.

5. **Tighten docs and skill guidance for OOB mode**

If we keep `--oob-host`, the docs and skills need to make the operational boundary clearer.

Please add/adjust guidance so that:
- Agents should never use `--oob-host` unless the user explicitly requests live/OOB callback testing.
- OOB substitution is generation-time behavior, not scenario YAML behavior.
- The scenario reference mentions live callback mode at least briefly and points to the adversarial payload reference.
- The validate skill/reference explains the OOB validation path once validation parity is added.
- The docs are clear that EvidenceForge writes payload text only; it does not execute the payload or make callbacks during generation.

6. **Keep the default inert behavior explicit**

The default canary behavior is important. Please keep making clear that, without `--oob-host`, payloads use the inert non-resolving canary and EvidenceForge does not execute or call out to anything during generation.

7. **Clean up docs wording around supported families and ground truth**

Once the proposed mechanism is removed, please make sure the docs consistently present all included families as supported.

Also please review wording around `GROUND_TRUTH.md` summaries for adversarial payloads. These values are not secrets, so showing escaped previews/full marked payload snippets is okay, but the docs should not imply secret-style redaction if the Markdown intentionally shows the payload content for analyst scoring.

I ran the focused adversarial-payload tests, `validate-config`, fixture validation, generation, eval, and targeted ruff checks locally. The core implementation looks solid overall; these requests are mostly about tightening the public contract now that we’re accepting the new capabilities.

peterhanily:
Thanks again for the thorough review, @DavidJBianco. I've pushed an update addressing all seven points, plus a few fixes and additional test coverage that came out of a deeper self-review. The branch is rebased onto current `dev`. Each point below is self-contained so you can spot-check it independently.

**Verification:** `uv run pytest --no-cov` is green (4,378 passed, 41 skipped); `ruff check` and `ruff format --check` are clean; `eforge validate-config` reports 0 errors across all config files. No new runtime dependency and no vendored corpus (`pyproject.toml` / `uv.lock` unchanged).

## Your review items

1. **Remove the `proposed` mechanism — done.** The `proposed` field is gone from the `PayloadFamilyEntry` schema (which is `extra="forbid"`, so a stray `proposed:` key is now rejected), the four `proposed: true` entries and "pending sign-off" wording are removed from `payload_families.yaml`, the validation warning is deleted, and the docs present all eight families as supported. A test asserts the full eight-family set is present.

2. **Tighten `--oob-host` — done.** It now requires a concrete registrable domain or an IP literal; bare TLDs / single labels (`com`, `fun`, `local`) and multi-label public suffixes are rejected, and the allowlist suffix-match stays scoped strictly under the accepted domain. Registrable-domain detection reuses the in-repo curated public-suffix list (no new dependency, per your guidance), which I expanded (~12 → ~38) to cover common ICANN ccTLD second-levels and vendor namespaces (`github.io`, `herokuapp.com`, `ngrok.io`, `s3.amazonaws.com`, …). A bare suffix is rejected while a name under it (`abc.github.io`) is accepted. Hostnames are now validated per-label (no empty/hyphen-edge labels, no all-numeric pseudo-IPs). It is a curated common subset, not the full PSL — documented as such, and overlay-extensible. Regression tests cover every case you enumerated.

3. **Remove `--i-am-authorized` — done.** Removed; passing `--oob-host` is now the sole explicit opt-in, and the loud `LIVE CALLBACK MODE` banner is retained.

4. **`eforge validate --oob-host` parity — done.** `validate` now accepts `--oob-host` (no auth flag, matching your example), normalizes/validates it through the same helper as `generate`, and threads the hosts into `ScenarioValidator` — so a literal-OOB scenario validates identically under both commands.

5. **Tighten OOB docs/skill guidance — done.** The `scenario`, `generate`, and `validate` skills and the reference docs now state: agents must not use `--oob-host` unless explicitly asked; OOB is generation-time only (never a scenario-YAML field); EvidenceForge writes payload text only and never executes it or calls out; the scenario reference mentions live-callback mode and links to the adversarial-payload reference; and the validate skill documents the OOB path.

6. **Keep default-inert behavior explicit — done.** The docs state in several places that, without `--oob-host`, payloads use the inert non-resolving canary (`canary.eforge.invalid`) and nothing is executed or contacted during generation; the code default is `oob_hosts=()` throughout.

7. **Docs wording (families + ground truth) — done.** All families read as supported. The `GROUND_TRUTH` adversarial-payload preview intentionally shows the full payload value — these are inert, marked test artifacts, not secrets — with control-byte / table-delimiter escaping only for safe rendering. The docs now say this explicitly rather than implying secret-style redaction.

## Beyond your review (from a deeper self-review)

I ran an empirical pass — generating data and inspecting the on-disk bytes — and found a couple of latent items worth flagging:

- **Phantom-positive validator bug (fixed).** The Linux-only-surface check gated on `os_cat == "windows"`, but a non-Linux/non-Windows host (macOS/BSD/Solaris → `unknown`) would pass validation and then be silently dropped at emit, producing a ground-truthed-but-never-emitted label. It now gates on `!= "linux"`, and `generate` aborts on the error. The identical bug in the `spillage` validator is fixed in a separate, standalone PR — #323 — so it can ship independently of this feature.

- **On-wire IDS labeling on URL surfaces (resolved docs-only — your call welcome).** On `http_request_url` / `http_referrer` the payload is percent-encoded on the wire, so the literal content token isn't byte-present and the `2012887` signature reads "CRLF in HTTP **Header**" even though the payload rode the URL query. The firing model you accepted is internally consistent (the recorded `ids_alert` always matches `snort_alert.log` on disk), so I deliberately left the behavior unchanged and instead documented the sensor model: it represents a URI/header-normalizing sensor, a raw-content rule may not fire, and the ground-truth `surface` field disambiguates where the payload actually rode. If you'd prefer the alert to be surface-gated (so a header-named rule does not fire for a URL-borne payload), that's a small follow-up and I'm happy to implement it whichever way you prefer.

- **Docs accuracy.** Clarified that `rendered_value` / `rendered_sha256` equals the whole on-disk field only for `syslog_message` and `http_user_agent`; for the URL surfaces (and `process_command_line`) it is a carrier-wrapped substring, so it should be hashed as a fragment.

- **Test coverage.** Added an all-eight-family × surface "lands on disk" matrix (the fixture exercised only three families), the `emitted: false` skip/no-leak invariant, the non-Linux OS-gate regression, and the `--oob-host` contract cases. I also realigned one on-the-wire Zeek test after rebasing onto current `dev`, where Zeek output now correctly requires a sensor on the path.

## Constraints honored

The family set and the public-suffix list remain small, curated, data-driven, and overlay-extensible, with no third-party corpus and no new dependency.

The one place I'd specifically value your preference is the IDS sensor-model question above. Otherwise I'm glad to adjust anything further — thanks for the careful review.

DavidJBianco:
Thanks for the update. I explicitly re-tested the two remaining concerns against PR head `496a7c61`.

Required fixes before merge:

1. Please enforce the OOB host safety contract closer to where OOB hosts become part of the payload allowlist, not only at CLI argument parsing.

The CLI now correctly rejects broad values like `--oob-host com`, but that validation happens before the adversarial-payload safety check. The lower-level safety path still trusts any `oob_hosts` value it receives and adds it to the host allowlist. If `com` reaches that path, a payload containing `evil.com` is accepted because `evil.com` suffix-matches `.com`.

The concern is not that programmatic generation is a supported public API; it is that the safety invariant lives too far out at one caller boundary. Please move/reuse the OOB host normalization/validation at the adversarial-payload safety boundary, or introduce a small validated OOB allowlist helper that `check_payload_safety()` / `resolve_value()` cannot bypass. The CLI can still call the same helper for fail-fast UX. Please add a regression test for broad values like `com` / `co.uk` reaching the safety path.

2. Please keep `expected_sources` limited to sources that actually land in generated output.

For a cleartext HTTP adversarial payload with only `web_access` configured, `GROUND_TRUTH.json` currently records:

```text
expected_sources= ['web_access', 'zeek_http']
```

But the generated output contains no Zeek HTTP file:

```text
WEB-01.example.com/web_access.log
```

`zeek_http` is the concrete reproducer here because it is appended as a secondary source based on `scheme == "http"`, not based on configured/observable output. More generally, `expected_sources` should mean “this source is expected to exist in this generated dataset,” not “this payload would theoretically be visible there.” Please gate secondary expected sources on actual configured/observable output, or split theoretical visibility into a separate field.

Everything else I spot-checked looks aligned with the requested design: typed event, deterministic generation, data-driven families, inert default behavior, no hidden callbacks/LLM/API calls, and strong final-output tests.

peterhanily:
Both required fixes are pushed (head `47a1fa58`) — details below so you can re-test directly.

**1. The OOB host contract is now enforced at the safety boundary, not only at the CLI.**

I moved the OOB-host validation into a single `normalize_oob_host()` helper in
`generation/adversarial_payload.py` and call it inside `check_payload_safety()` itself — which
`resolve_value()` and `ScenarioValidator` already funnel through. A broad value handed straight
to the safety path (not via the CLI) is now rejected before it can enter the host allowlist:
`check_payload_safety(value, oob_hosts=("com",))` and `resolve_value(..., oob_hosts=("com",))`
both raise, so a payload containing `evil.com` can no longer suffix-match `.com`. The CLI's
`_normalize_oob_hosts` now calls the same helper purely for fail-fast UX, and the defense-in-depth
carrier-host check in `render_for_surface()` uses it too — so every layer that builds the
allowlist applies one identical contract. A bare TLD / single label / public suffix is rejected;
a concrete registrable domain or IP literal (and a subdomain of one) is accepted.

Regression test added per your note: bare TLDs and public suffixes (`com`, `fun`, `local`,
`co.uk`, `github.io`) passed directly to `check_payload_safety` / `resolve_value` now raise, while
a concrete registrable domain or IP literal still passes.

**2. `expected_sources` is now limited to sources that actually land in the output.**

`zeek_http` is no longer appended on `scheme == "http"` alone. It is added only when (a) the
`zeek_http` emitter is actually configured for the run and (b) a network sensor on the
connection's path observes Zeek — i.e. only when the row will exist on disk. For your reproducer
(cleartext HTTP with only `web_access` configured), `expected_sources` is now `['web_access']`
and no Zeek file is claimed. With a sensor that captures Zeek for that path, `zeek_http` is
recorded and the `http.json` row is present. Regression tests cover both directions, plus a
matrix-wide check that every entry in `expected_sources` corresponds to a file actually produced.

I re-ran the focused adversarial-payload tests, `validate-config`, fixture validate/generate/eval,
and ruff — all green; generation remains byte-for-byte deterministic across runs.

Thanks for the testing notes — testing the invariant at its owning boundary, and checking
`expected_sources` against the actual generated output, are now part of the regression suite.
Happy to adjust anything further.

DavidJBianco:
Thanks for the substantial contribution here, @peterhanily. The feature is well scoped for the log-pipeline hardening use case, and the follow-up fixes addressed the safety-boundary and ground-truth concerns cleanly. I’m going to merge this into `dev` and include it in the existing `dev` → `main` PR with an updated changelog.

PR #297, opened 2026-06-05
fix: render file_read side effects instead of crashing on action 'read'

peterhanily:
## Summary

A process-driven file side-effect whose action is `read` crashes generation mid-run with `KeyError: 'read'`.

The activity generator maps a file side-effect's `action` to an event type in two places, but both maps only handled `create` / `modify` / `delete`. Meanwhile `config/activity/edr_pools.yaml`'s `file_side_effect_profiles` can emit a `read` action — for example `linux_sshd_listener_state`, where `sshd` reads `/etc/ssh/sshd_config` — so when baseline activity selects that side-effect, the lookup raises `KeyError: 'read'` and aborts the run.

`file_read` is already a fully-modeled event type (`FileContext` allows `action="read"`, and the eCAR emitter renders it as a `FILE` / `READ` record), so the read side-effect should simply be emitted rather than crashing.

## Fix

- Map `read` → `file_read`, so a file-read side-effect renders as a normal eCAR `FILE READ` event.
- Consolidate the two byte-identical inline maps into a single module-level `_FILE_EFFECT_EVENT_TYPE` constant.
- Add a regression test asserting the map covers every action the EDR config can emit, and that each maps to an event type the eCAR emitter renders.

## Notes

The crash was non-deterministic in practice: it only triggered when baseline activity happened to select an `sshd` file side-effect (more likely under sysadmin-persona / medium+ intensity). After the fix, a previously-crashing scenario generates cleanly and the `sshd_config` `FILE`/`READ` events render correctly.

`pytest --no-cov`, `ruff check`, and `ruff format --check` pass.

### DIFF
--- a/src/evidenceforge/generation/activity/generator.py
+++ b/src/evidenceforge/generation/activity/generator.py
@@ -236,6 +236,18 @@ class _HttpPersistentConnection:
 
 _HTTP_PERSISTENT_REUSE_GUARD = timedelta(milliseconds=900)
 
+# Map a process-driven file side-effect `action` (from edr_pools.yaml's
+# file_side_effect_profiles + the command-semantic effects) to its canonical event
+# type. Must cover EVERY action those data sources can emit — a missing key is a
+# generation crash (e.g. the sshd `read` of /etc/ssh/sshd_config). `file_read` is a
+# fully-modeled event (FileContext action "read", eCAR FILE/READ).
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Closing as superseded by 7e3a3a52, which already fixed the runtime crash by adding the read -> file_read mapping on current dev. We carried forward the useful regression coverage separately in 4cf50430 against current dev.

PR #311, opened 2026-06-17
fix: reject symlinked SOF-ELK staged logs

DavidJBianco:
### Motivation

- Prevent accidental disclosure of local files when staging discovered logs for SOF-ELK by addressing an unsafe copy pattern that followed symlinks and allowed resolved targets outside the generated output root.

### Description

- Add `_safe_stage_source(source_root, source)` which rejects symlinked discovered files and verifies `source.resolve()` is inside `source_root` before staging.
- Use the safe resolved path in `stage_source_logs()` for year inference and the `shutil.copyfile()` source argument while preserving the original discovered `source` for metadata such as the sensor name.
- Update logic so year partitioning still works (compute `source_year` from the validated resolved path for `syslog` families).
- Add regression tests in `tests/unit/test_sof_elk_sources_harness.py` to assert that symlinked `web_access.log` files are rejected and that sources resolving outside the generated root are rejected.

### Testing

- Ran lint/format checks: `uv run ruff check .` and `uv run ruff format --check .`, both passed after autofix.
- Ran targeted unit tests: `uv run pytest --no-cov tests/unit/test_sof_elk_sources_harness.py` and `uv run pytest --no-cov tests/unit/test_external_parser_runner.py tests/unit/test_sof_elk_sources_harness.py`, both passed.
- Ran full test suite: `uv run pytest --no-cov` which completed successfully (3943 passed, 40 skipped in this environment).
- Verified the new tests fail on the vulnerable behavior and pass with the fix implemented.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a203107d43883258fe504139116ecda)

### DIFF
--- a/src/evidenceforge/external_parsers/sof_elk_sources.py
+++ b/src/evidenceforge/external_parsers/sof_elk_sources.py
@@ -282,14 +282,15 @@ def stage_source_logs(
     logs: list[StagedSourceLog] = []
     for source_name in spec.source_names:
         for source in sorted(source_root.rglob(source_name)):
+            safe_source = _safe_stage_source(source_root, source)
             sensor = _source_name(source_root, source)
-            source_year = _source_year(source) if spec.staged_directory == "syslog" else None
+            source_year = _source_year(safe_source) if spec.staged_directory == "syslog" else None
             if spec.staged_directory == "syslog" and source_year is not None:
                 destination = source_stage_root / str(source_year) / sensor / spec.staged_name
             else:
                 destination = source_stage_root / sensor / spec.staged_name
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Rejected as superseded by #319. The issue is real, but #319 addresses the same external-parser source-staging symlink class at the stronger layer: it validates staged source paths, rejects symlink/non-file/out-of-root sources, and copies regular files with O_NOFOLLOW/fstat protection. I merged #319 into dev with the accepted Codex batch instead of taking this narrower staging-only fix.

PR #318, opened 2026-06-17
fix: reject symlinked external parser sources

DavidJBianco:
### Motivation
- The syslog SOF-ELK source path used a plain `shutil.copyfile(source, destination)` on files discovered with `rglob`, which follows symlinks on POSIX and allowed attacker-supplied `syslog.log` symlinks to cause out-of-tree file copies.
- Detection also resolved and used symlinked paths which could crash or misclassify a supplied data directory when a symlink pointed outside the `data/` tree.

### Description
- Added `_validate_stage_source(source_root, source)` and call it from `stage_source_logs` to reject symlinked source files and any resolved source path that is not contained under `source_root`, raising `SofElkHarnessError` on violation (prevents `shutil.copyfile` from following symlinks). (`src/evidenceforge/external_parsers/sof_elk_sources.py`)
- Hardened detection in `_add_detected_log` to skip symlink paths and to ignore resolved paths that are not under `data_dir`. (`src/evidenceforge/external_parsers/runner.py`)
- Tightened `_candidate_log_files` to skip symlinks and only return resolved candidate files that are contained inside the `data_dir`. (`src/evidenceforge/external_parsers/runner.py`)
- Added regression tests: `test_stage_source_logs_rejects_symlink_sources` and `test_detect_external_parser_plan_ignores_symlinked_supported_logs` to verify staging rejects symlinks and detection ignores them. (`tests/unit/test_sof_elk_sources_harness.py`, `tests/unit/test_external_parser_runner.py`)

### Testing
- Ran unit tests with `uv run pytest --no-cov tests/unit/test_sof_elk_sources_harness.py tests/unit/test_external_parser_runner.py` and all tests passed (`23 passed`).
- Ran linters/format checks with `uv run ruff check .` and `uv run ruff format --check .`, and both checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a203106ea188325b90669a0b9a576bd)

### DIFF
--- a/src/evidenceforge/external_parsers/runner.py
+++ b/src/evidenceforge/external_parsers/runner.py
@@ -260,7 +260,11 @@ def _add_detected_log(
     validator: str | None,
     unsupported_reason: str | None = None,
 ) -> None:
+    if path.is_symlink():
+        return
     path = path.resolve()
+    if not path.is_relative_to(data_dir):
+        return
     logs_by_path[path] = DetectedLog(
         path=path,
         host=_host_for_path(data_dir, path),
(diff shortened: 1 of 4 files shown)


DavidJBianco:
Rejected as superseded by #319. The issue is real, but this branch is a narrower overlap with #319. The accepted #319 fix covers source staging and discovery while adding stronger final-path protections via regular-file validation and O_NOFOLLOW copy, so merging both would duplicate and partially conflict in the same external parser files.

PR #320, opened 2026-06-17
fix: skip out-of-tree external parser logs

DavidJBianco:
### Motivation

- Automatic external-parser discovery resolved symlink targets and could pass paths outside the supplied `data_dir` into `_host_for_path`, which raised an uncaught `ValueError` and aborted validation.

### Description

- Add `_resolve_log_file_in_data_dir(data_dir, path)` that resolves a path and rejects it if the resolved target is not contained under `data_dir`.
- Update `_candidate_log_files` to only return resolved candidates that pass the containment check instead of blindly returning `path.resolve()` results.
- Update `_add_detected_log` to use the containment helper and skip discovered files whose resolved target is outside `data_dir`.
- Add regression tests `test_detect_external_parser_plan_skips_symlinked_logs_outside_data_dir` and `test_detect_external_parser_plan_skips_named_symlinked_logs_outside_data_dir` to cover symlinked files pointing outside the data directory.

### Testing

- Ran `uv run pytest --no-cov tests/unit/test_external_parser_runner.py` and the test file passed (6 passed). 
- Ran linter/format checks with `uv run ruff check .` and `uv run ruff format --check .` and both checks passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a203106bf248325a1ee1577e52ac50f)

### DIFF
--- a/src/evidenceforge/external_parsers/runner.py
+++ b/src/evidenceforge/external_parsers/runner.py
@@ -260,10 +260,13 @@ def _add_detected_log(
     validator: str | None,
     unsupported_reason: str | None = None,
 ) -> None:
-    path = path.resolve()
-    logs_by_path[path] = DetectedLog(
-        path=path,
-        host=_host_for_path(data_dir, path),
+    resolved_path = _resolve_log_file_in_data_dir(data_dir, path)
+    if resolved_path is None:
+        return
+
+    logs_by_path[resolved_path] = DetectedLog(
+        path=resolved_path,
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Rejected as superseded by #319. The issue is real, but this branch only handles out-of-tree symlink discovery in the external-parser runner. The accepted #319 fix covers that discovery class plus staging-time source validation and safe regular-file copying, so #320 is no longer the best root-cause fix to merge.

PR #323, opened 2026-06-17
fix: reject Linux-only spillage surfaces on any non-Linux host

peterhanily:
## Summary

The `spillage` validator (added in #289) has a latent phantom-positive bug — the same one I hit while hardening the sibling `adversarial_payload` validator in #296. The Linux-only-surface check gates on `os_cat == "windows"`, but `_get_os_category` returns `windows` / `linux` / **`unknown`**.

As a result, a `shell_history` or `syslog_message` spill placed on a host whose OS is neither Windows nor Linux — e.g. macOS, FreeBSD, Solaris, AIX (all classified `unknown`) — passes `eforge validate` cleanly, yet the emitter (which only models `linux`) silently drops it at generation. The outcome is a **ground-truthed-but-never-emitted label**: `GROUND_TRUTH` claims the credential is present, but it never lands on disk — a guaranteed false positive for any scrubber/DLP scored against the dataset, with no signal to the author that anything went wrong.

## Fix

Gate the check on `os_cat != "linux"` (the emitter only models `linux`) and report the detected OS category in the error message. `eforge generate` already aborts on a validation error, so the phantom positive is blocked end to end.

## Notes

- Found while addressing the review on #296; split into this standalone PR so the fix to already-released code can ship independently of that larger feature.
- Regression test added in `tests/integration/test_spillage_generation.py`, parametrized over macOS / FreeBSD / Windows.
- `uv run pytest --no-cov`, `ruff check`, and `ruff format --check` all pass.

### DIFF
--- a/src/evidenceforge/validation/schema.py
+++ b/src/evidenceforge/validation/schema.py
@@ -1235,19 +1235,21 @@ class ScenarioValidator:
                     continue
                 field_path = f"storyline.{idx}.events.{spec_idx}"
 
-                # shell_history/syslog are Linux-modeled; emitting them on Windows
-                # would silently drop the credential while still labeling it in
-                # ground truth — a phantom positive — so this is a hard error.
-                # (process_command_line is cross-OS, so it is exempt.)
-                if os_cat == "windows" and spec.surface in LINUX_ONLY_SURFACES:
+                # shell_history/syslog are Linux-modeled; emitting them on any non-Linux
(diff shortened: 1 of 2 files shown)


DavidJBianco:
Accepted via a rebased/amended commit on dev: de6fc246. I kept the root-cause validator change, expanded the regression to cover both Linux-only spillage surfaces (shell_history and syslog_message) across macOS, FreeBSD, and Windows, updated the v1.3.3 changelog, and included it in release PR #328. Closing this stale PR so the accepted fix is tracked through dev.

PR #333, opened 2026-06-18
feat: add LLM prompt-injection demo scenario (closes #285)

### LINKED ISSUE #285: PR Proposal: Prompt Injection Demo Scenario
The CONTRIBUTING guide lists "Create new example scenarios" under Other Ways to Contribute

LLM-based SOC copilots are vulnerable to indirect prompt injection delivered
through untrusted log content. There is no public synthetic dataset demonstrating this.

I'd like to contribute a single example scenario under
`scenarios/llm-injection-demo/` that uses the existing `RawEventSpec` to
plant ~12 prompt-injection payloads (across 5 Watchtower attack classes)
into HTTP user-agent, syslog MSG, and DNS QNAME fields of an otherwise
ordinary scenario.

  - No new event types, no schema changes, no new CLI commands.
  - Just YAML + a README explaining what defenders should test their LLM
    copilots for.
  - Every payload tagged with `EFORGE-CANARY-<uuid>` for scoping.
  - `GROUND_TRUTH.md` generated normally by `eforge generate`.

I have a tool https://forensicate.ai/ that can generate these but there's many others and these get stale quickly so don't want to make the scenario so narrow it'll become dated fast. 

Any concerns with the topic before I open the PR? 

Thanks,

Peter

peterhanily:
Closes #285.

## What this adds

A single example dataset — `scenarios/llm-injection-demo/` — for testing whether an **LLM SOC copilot** (an assistant that triages/summarizes logs) can be hijacked by **indirect prompt injection delivered through untrusted log content**. Many log fields are attacker-controlled (a User-Agent, a URL, a DNS query, a failed-logon username), and a copilot that reads them has no enforced boundary between untrusted *data* and trusted *instructions*. This is OWASP **LLM01:2025** (the indirect variant) / MITRE ATLAS **AML.T0051.001**, and there was no public synthetic dataset demonstrating it.

Concretely:

- **The scenario** — `scenario.yaml`, a matched `scenario-clean.yaml` twin, and a `README.md` with the self-scoring recipe: an ordinary small-enterprise dataset with **14 labeled prompt-injection payloads** and **2 decoy credentials** planted in it.
- **4 payload families** (curated data in `config/activity/payload_families.yaml`): `prompt_injection_persona` (authority/persona hijack), `prompt_injection_context` (forging a log/markup boundary so the copilot reads a fake "BENIGN" conclusion as its own — includes the guardrail / parser-shutdown variants), `prompt_injection_exfil` (data-exfiltration / tool-call abuse), and `prompt_injection_control` — a **labeled negative control** (a blunt "ignore previous instructions" plus obfuscations) that a hardened copilot should resist; emission there means a trivially-broken reader.
- **2 new `adversarial_payload` surfaces**: `dns_qname` (the payload is a DNS query name, read in Zeek `dns.log`) and `auth_user` (the payload is a failed-logon username, read in `auth.log`). Together with the existing User-Agent / request-URL / Referer / syslog surfaces, the demo exercises the attacker-controlled fields a copilot actually reads.

## How the payloads reach the logs (the design question on #285)

Not raw events. The payloads are emitted through the existing typed `adversarial_payload` event, so each injection flows through the canonical generation path and correlates with the surrounding evidence — the request/connection/auth records, the Zeek rows, and `GROUND_TRUTH.json` — instead of being a disconnected raw line. The proposal called for a DNS-QNAME surface; delivering it (and `auth_user`) meant adding those two surfaces to that event's surface set. Both render through the canonical DNS-lookup and failed-logon paths, not bespoke emitter code.

## Proof without a callback (the proof/flag-file question on #285)

Every injection embeds a unique **echo-canary** — `EFORGE_TEST-CANARY-<nonce>` — that it instructs the copilot to emit *in its own verdict*. A defender greps their copilot's output for it: deterministic, callback-free, model-agnostic. There is deliberately **no flag-file write** (that would require executing the payload) — the canary is the proof. And because generation is deterministic, the matched `scenario-clean.yaml` twin shares a **byte-identical baseline** with every payload neutralized, so a defender can diff their copilot's verdicts across the two runs to isolate the injection — including silent spillover onto neighbouring alerts that carry no token of their own.

## Inert by construction

- Every payload carries the `EFORGE_TEST` marker on **every physical line**.
- Any host that appears is the non-resolving canary (`canary.eforge.invalid`, RFC 6761) or an RFC-reserved address — re-checked at config load **and** at generation. A real registrable domain is rejected; on the `dns_qname` surface (where the payload *is* a hostname) the query name always renders under the reserved canary domain.
- Payloads are written into log fields as text and never executed; generation runs no model and makes no network calls. The default run uses the inert canary; an opt-in `--oob-host` mode exists only for operators who explicitly want a live out-of-band test.
- **No new dependencies** — the families are curated in-repo data.

## Staying current

Prompt-injection *techniques* date fast; the *structure* does not. The durable parts — the attack classes, the data/instruction trust-boundary failure, and the echo-canary proof — live in each family's `weakness_class` text and the README. The volatile parts — exact wordings, encodings, and any cited figures — live as `value_templates` and can be refreshed (or extended per-model via a project-local overlay) with no code change.

## Ground-truth integrity, validation, and tests

- `eforge validate` / `validate-config` clean; `eforge generate` + `eforge eval` clean (Event Presence 100/100 on the demo).
- **Ground truth equals what lands on disk.** A payload that can't be recorded is labeled `emitted: false`, never a phantom-positive label. The new surfaces fail fast in validation when they can't land: `dns_qname` requires a network sensor emitting Zeek (a host keeps no DNS log of its own), and `auth_user` is Linux-modeled.
- Unit + integration tests cover both new surfaces across OS categories (Linux-only surfaces rejected on macOS/FreeBSD/Windows), `expected_sources` ⇔ files-actually-produced, the host-safety boundary, byte-for-byte determinism, and the demo end-to-end (canary recovery, decoy landing, negative-control labeling, and the byte-identical twin).
- Reference docs (`docs/reference/adversarial_payload.md`, `docs/reference/scenario-reference.md`) and the `/eforge` skills are updated for the two new surfaces.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DavidJBianco:
Thanks for the contribution, @peterhanily. This is a strong addition: the demo lands the prompt-injection payloads through modeled `adversarial_payload` surfaces rather than disconnected raw rows, and the echo-canary/clean-twin scoring workflow gives defenders a practical way to test SOC copilots.

I made one small README wording tweak before merging: the clean twin now describes the comparison as sharing the same baseline activity and almost all source-native rows, with expected differences limited to injected fields and directly related transport rows. That matches what we observed when generating and diffing the poisoned and clean datasets.

Validation run locally on the final PR head (`5009700d`):

```bash
uv run --extra dev pytest --include-slow --no-cov
```

Result: `4475 passed, 28 skipped`.

PR #389, opened 2026-08-16
feat: labeled dual-signal SnortML+signature and Splunk-notable triage corpus

AAH20:
## Summary
Adds a labeled dual-signal triage corpus for SnortML-style (GID:411) scores,
classic signature alerts, and Splunk-notable-shaped events.

## Why
SnortML documentation and agentic-SOC practice require that ML probability
not be treated as equivalent to a signature true positive. Training corpora
that flatten alerts into a single "bad" label teach the wrong automation habit
and inflate false containment and agent compute burn.

## What's included
- Labeled events: signature-only, ML-only high, corroborated, ML-low, Splunk notables
- Disposition labels suitable for T1-T3 automation evaluation
- Validator that rejects corpora marking ML-only high as `auto_contain` or `fix_now`
- Unit tests and README scenario entry

## Out of scope
Engine parser CVEs, which belong in Cisco PSIRT rather than this PR.

## Production application
Aegis Decision Fabric exercises this corpus for composite-confidence evaluation,
gated remediation, and FP/TP feedback:
https://github.com/AAH20/aegis-decision-fabric

Secure Firewall and Splunk teams can review the fixed-scope implementation offer
at https://a2zsoc.com/consultation.

## Test plan
- [x] `python scripts/validate_dual_signal_triage_corpus.py scenarios/dual-signal-triage-corpus/corpus/labeled_events.json`
- [x] `pytest tests/unit/test_dual_signal_triage_corpus.py --no-cov`
- [ ] Maintainer CI green on this branch


DavidJBianco:
Thanks for the contribution. I’m having trouble understanding how this corpus is intended to relate to EvidenceForge’s functionality and purpose.

As I understand the PR, each prescribed disposition is based on a single alert-like JSON record containing fields such as the signal class, GID/SID, ML score, severity, and source/destination addresses. There does not appear to be any surrounding evidence, such as endpoint activity, network connections, authentication events, process execution, asset or user context, related alerts, or scenario ground truth, from which an analyst or agent could actually triage/judge the alert.

That raises a few questions:

- Is this intended to test alert-routing policy, actual alert triage, or something else? The current cases appear to prescribe a disposition directly from the alert type. Real triage or analysis would require supporting evidence.
- Are these records intended to correspond to EvidenceForge scenarios that generate the underlying activity and logs from which these alerts would be produced? If so, where are those scenarios, and how would the corpus records be correlated with the generated evidence and scenario ground truth?
- What is the intended end-to-end EvidenceForge workflow? The README suggests pairing the corpus with EvidenceForge-generated data, but the PR does not seem to provide a pairing mechanism, identifiers, example scenario, or evaluation procedure.

Could you clarify the intended integration and provide a concrete end-to-end example showing how this contribution would be used with EvidenceForge? I want to be sure that I actually understand the value of what you're proposing to contribute.

AAH20:
You're right, and thanks for pushing on it. I dug into what was actually here: no underlying evidence, no logs, no scenario ground truth -- just alert metadata with a pre-assigned disposition and a validator that only checked internal self-consistency. That's a fair thing to flag.

Fixed it rather than just explaining it away. New commit adds `evidence-scenario.yaml`, an EvidenceForge scenario (generated via `eforge`, seed 1234) that produces the real underlying activity each labeled event now points to:

- **signature_only_high / signature_plus_ml_corroboration**: real SQLi HTTP requests that actually fired the Emerging Threats signature (SID 2009714) -- confirmed in `snort_alert.log`, not asserted.
- **snortml_gid411_high_ml_only**: a real SQLi request (error-based `extractvalue`) that the same curated signature set does **not** match. That's a genuine detection blind spot the tool models on purpose, not an invented absence of evidence -- exactly the "ML catches what signature-only misses" case this corpus is about.
- **snortml_low**: an ordinary, unremarkable product-catalog request. Nothing malicious in it, which is the point.
- **splunk_notable_high_risk / _low**: real Windows 4625 failed-logon volume (61 attempts vs. 2), not a hand-picked risk score.

More importantly, `validate_dual_signal_triage_corpus.py` no longer only checks that the corpus agrees with itself. It now opens the generated files and verifies each event's `evidence.sources[]` match string is actually there, and cross-checks `ids_alert` presence in `GROUND_TRUTH.json` against what each event's `signal_class` claims -- so an `ml_only` label pointing at a request that actually fired the signature fails validation instead of shipping. I verified this catches real regressions before pushing (tampered-reference test in the new test file).

To answer your three questions directly: this isn't yet paired with a full attack-scenario storyline (no lateral movement, no broader narrative) -- it's scoped to the six triage cases with real, targeted evidence behind each. If you want it extended into a fuller scenario, that's a reasonable next step and `evidence-scenario.yaml` is a real starting point for it now rather than a placeholder.

DavidJBianco:
Thanks for clarifying, but this still seems like it's adjacent to but not part of EvidenceForge's intent. We're producing realistic logs, but I think evaluating a triage system is out of our scope. I think this is useful, but would be more appropriate as a separate offering, maybe that relies on EvidenceForge for generation.

